PR #2, opened 2023-01-26
Restructure directories

maxwolfs:
* update folder structure
* fix dead links and markdown formatting


itrich (commented):
In the certification YAML, we're using `kaas` instead of `container`. We should IMHO stick to one naming scheme.

fkr:
> In the certification YAML, we're using `kaas` instead of `container`. We should IMHO stick to one naming scheme.

full ack. Thanks for pointing that out @itrich!

itrich (approved):
LGTM!

PR #8, opened 2023-02-11
Add kaas project

jschoone:
Hi,
please provide me a project so I can do tests on and with the k8s-cluster-api-provider. I tried doing tests in another projects but went in conflicts because of already existing resources which is a blocker when doing full tests.

Signed-off-by: Jan Schoone <jan@jass.es>


frosty-geek (approved):
LGTM, project created and should be usable now

PR #17, opened 2023-03-03
Fix broken link after restructuring

maxwolfs:


berendt:
In my time in the OpenStack documentation team I once built a script that searches for dead links in the documentation. This could be useful for us to use. I'll see if it still exists somewhere.

itrich:
:+1: for a GitHub Action workflow that looks for dead links and creates an issue (similar to the workflow in our github-manager). There are various free web services for finding 404 services that also offer an API (and probably there are already existing GitHub Actions for this use case).

berendt:
Have only found my sitemap generator (https://github.com/openstack/openstack-doc-tools/tree/master/sitemap).

I think I had done that 404 check with Scrapy at the time. You should be able to do this quickly. Unfortunately, I can not find the script anymore.

maxwolfs:
Good Point and nice idea which would help a lot! I can have a look at it or should we do it together @itrich ?

Could you approve for merge. Currently blocks 'doc-page' build. Merci!

fkr (approved):
LGTM.

fkr:
I very much concur with @itrich and @garloff that we need a static link service, so that we can persist certain links in a good way (see https://github.com/SovereignCloudStack/standards/pull/249 for that conversation)

PR #22, opened 2023-03-09
Fix minor issues in contribute docs

joshmue:
Just minor things I'd consider to adjust. All in all, after reading this document, that I feel like I do understand how the mechanism works, so thanks for sharing!

---

Regarding the changes to the inline JSON:

1. As plain JSON does not support comments (but this example of course benefits from comments), using `json5` as format of the code block results in a nicer visual rendering - even though it will not be valid plain JSON either way.
2. I'd usually expect comments to be above the thing that they are commenting on - not below.
3. (not touched by this PR:) The comments could be maybe a bit more precise regarding the type of data which is to be expected.
    - `link to github organisation and repository` has `demo-organisation/demo-repository` as example, which is not a link per se.
    - `directory which shall be copied. Optional to specify only copying markdown files` has `doc/*.md` as example, which is not a directory, but a glob which can expand to `n` files, which is probably what `Optional to specify only copying markdown files` says, but I did not get that first try.
    - `label for directory. only mandatory if source file is set to copy only *.md files and not the complete directory` kind of clarifies the previous point: It can be set to a directory as well as to a glob matching individual files.

### DIFF
--- a/community/contribute/adding-docs-guide.md
+++ b/community/contribute/adding-docs-guide.md
@@ -10,20 +10,20 @@ Determine the type of your documentation and click to continue.
 2. [Operational documentation](#2-operational-documentation)
 3. [Community documentation](#3-community-documentation)
 
-If unsure don't hestitate to ask us at [Matrix](/community/communication/matrix.md)
+If unsure don't hestitate to ask us at [Matrix](/community/communication/matrix.md).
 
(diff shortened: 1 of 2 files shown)


maxwolfs (approved):
Thank you very much for the great feedback and additions @joshmue!

PR #34, opened 2023-06-22
Add checklist for Hackathon planning

ra-beer:
This Patch provides a list for planning hackathons.

Signed-off-by: Ramona Beermann <ramona.beermann@osism.tech>

PR #36, opened 2023-07-03
Update .markdownlint-cli2.jsonc

maxwolfs:
* allow for inline-html but only `<input>` for allowing checkboxes within tables see, [here](https://github.com/SovereignCloudStack/docs/pull/34)


ra-beer (approved):
LGTM Thanks

PR #40, opened 2023-07-10
Add ansible-styleguide

ra-beer:
Signed-off-by: Ramona Beermann <ramona.beermann@osism.tech>

### DIFF
--- a/community/contribute/styleguides/ansible_styleguide.md
+++ b/community/contribute/styleguides/ansible_styleguide.md
@@ -1,0 +1,156 @@
+# Ansible Style Guide
+
+We use nearly all default rules of ansible lint. A listing of all these rules can be found in the Ansible Lint documentation:
+<https://ansible.readthedocs.io/projects/lint/rules/>.
+Please always use the ansible linting to check if the code complies with the default linting rules.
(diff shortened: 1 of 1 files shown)


community/contribute/styleguides/ansible_styleguide.md
```diff
@@ -0,0 +1,156 @@
...
+    version: main
+```
+
+## Usage of roles from other collections
+
+If you want to reuse roles please do it in the following way:
+
+First you have to declare the role or collection in the requirements.yml like shown in the example before.
+
+Than you can use it in plybooks like this
```
maxwolfs, line 122:
```suggestion
Than you can use it in playbooks like this
```

maxwolfs (changes requested):
I found one typo otherwise lgtm! 
One question though: Should we put `when` and `become` into code blocks to easier distinguish those parameters from text?

community/contribute/styleguides/ansible_styleguide.md
ra-beer: Typo is done. There are more parameters not only when and become. Like changed_when or collections. The most things in this styleguide describes how parameters should be used or placed. I think the example codeblocks are enough for visibility.

maxwolfs (approved):

PR #41, opened 2023-07-20
Add openstack projects for tender vp06a

jschoone:
Hi,
we need two more environments. One is for a tender to work on the Network and Storage Interfaces for Kubernetes. 
Thank you :cloud:


frosty-geek (approved):
created new project as requested /cc @jschoone

PR #42, opened 2023-07-20
Add openstack projects for e2e tests for kubernetes

jschoone:
We need to do some e2e tests for our k8s-cluster-api-provider using Zuul


jschoone:
Hi @linwalth, can you help here?

PR #50, opened 2023-08-17
:sparkles: Adding the cluster-stack-operator design document

batistein:
We're introducing a design for an operator aimed at enhancing cluster stack usage in Kubernetes. Building on Cluster-API advancements, this operator will streamline cluster operations and elevate user experience.

- **Purpose:** Facilitate streamlined use of cluster stacks in Kubernetes.
- **Integration with Cluster-API:** Builds upon Cluster-API advancements for effective cluster operations.
- **Component Bundling:** Combines fundamental cluster elements, configuration settings, and node images.
- **Release Monitoring:** Tracks and integrates new releases automatically.
- **Management Simplification:** Reduces complexity in cluster stack management tasks.
- **User Narratives:** Provides scenarios of anticipated user-system interactions.
- **History and Evolution:** Discusses cloud orchestration and cluster stack integration.
- **Risk Analysis:** Addresses potential challenges, especially for providers integrating the operator.
- **Repository Structure:** Detailed breakdown for better understanding and development.
- **Tool Integration:** Uses tools like kubebuilder for CRD validation.
- **Cluster Objects Exploration:** Deep dive into essential cluster objects and their roles.
- **Controller Reconcile Loops:** Focuses on precision in release monitoring, especially with auto-subscribe features.

[Click here for a preview](https://github.com/SovereignCloudStack/docs/blob/syself/cluster-stack-operator/operations/kaas/cluster-stack-operator.md)

### DIFF
--- a/operations/kaas/cluster-stack-operator.md
+++ b/operations/kaas/cluster-stack-operator.md
@@ -1,0 +1,692 @@
+# Cluster Stack Operator Design Document
+
+## Introduction
+
+This document describes the design of an operator that factilitates using the cluster stacks. The cluster stacks are described in a separate document.
+
+Our cluster stacks, leveraging the capabilities of Cluster-API, provide a significantly enhanced user experience. This enhancement is achieved by packaging together various essential components necessary for effective cluster operation. These components encompass core elements of a cluster, configuration settings, node images, and more.
+
+A key feature we utilize within the Cluster-API is the `ClusterClass`. The `ClusterClass` enables us to curate a comprehensive "cluster bundle". Within this bundle, we encompass all the vital components needed for a fully functional, production-ready cluster. This includes the kubeadm configurations and other crucial applications. Users can effortlessly create a custom cluster by simply referencing the desired `ClusterClass` object to the Cluster.
+
+In order to solve some issues that go beyond core CAPI components, we bundle application configuration, node images, and the `ClusterClasses` to formulate what we term as "cluster stacks".
+
+While the Cluster-API is adept at managing the lifecycle of clusters, there are aspects it doesn't cover. For instance, it doesn't oversee the lifecycle of the Cluster Stack components. For example, when a new Syself Cluster Stack is introduced, the corresponding `ClusterClass` object should be downloaded and implemented, ensuring it's readily accessible to users. Additionally, there are certain tasks that are provider-specific, such as building node images, that the Cluster-API doesn't cater to.
+
+This is where we want to introduce the cluster-stack-operator, which is supposed to handle all of these tasks that would otherwise require manual work of the user when using Cluster-API.
+
+
+## Motivation
+
+The Cluster Stacks contain the configurations of many different applications, e.g. Cilium. Additionally, they contain a specific node image. This node image can be updated with a new version of a Cluster Stack, but does not have to.
+
+If there is a new version of Cilium, this should be made available for users of any Cluster Stack that contains Cilium. Assuming that there is a `ClusterStackXYZ` with Cilium. After changing the respective versions in the configuration, a new release of `ClusterStackXYZ` will be published. Users that have clusters that point to the `ClusterClass` object of `ClusterStackXYZ` want to update to the latest release. This has to be done by first applying some yaml files in the management cluster (e.g. the `ClusterClass` object). Then, the node image has to be built if it doesn’t exist already.
+
+This process is a) provider-specific and b) contains multiple steps of manual work.
+
+## Proposal
+
+Introducing the cluster-stack-operator: a transformative tool designed to alleviate users from the painstaking manual tasks traditionally associated with the usage of cluster stacks. The primary objective of this operator is to streamline and automate the following pivotal processes:
+
+1. **Proactive Release Monitoring:** The operator will be vigilant, regularly scanning for the newest releases of cluster stacks, ensuring users are always updated with the latest enhancements and features.
+
+2. **Streamlined Release Preparation:** Once a user expresses interest in a particular cluster stack release, the operator jumps into action, preparing the release for smooth deployment and use.
+
+3. **Efficient Clean-Up:** Over time, some cluster stack releases may become obsolete or redundant. The operator will diligently purge unused and outdated releases, ensuring the system remains clutter-free and optimized.
+
+A noteworthy feature of our operator is its modular architecture. Given that some aspects of its operation, such as building node images, are provider-specific, we've strategically adopted the structural framework utilized by the Cluster-API. This means that provider-specific tasks are delegated to a distinct operator. This specialized operator then synergizes with the core operator to prepare cluster stack releases for deployment.
+
+In a bid to offer users an intuitive experience and seamless transition, we've ensured that our object structure mirrors that of the Cluster-API. This alignment not only facilitates user understanding but also fosters a harmonious integration between the two closely related projects.
+
+However, there is a key distinction between our operator and the Cluster-API. Not all providers necessitate provider-specific work. Recognizing this, we've imbued our system with flexibility. There isn’t a compulsory need to construct a provider-specific operator. Instead, users have the autonomy to determine whether they wish to initiate such an operator based on their unique requirements. This approach affirms our commitment to offering a user-centric tool, tailored to meet diverse needs with precision.
+
+## User stories
+
+### User Story 1: Self-service use of SCS cluster stacks by Human Operators
+
+As a human operator, I desire the ability to harness the capabilities of the SCS cluster stacks in a self-service mode. In doing so, I expect the cluster-stack-operator to guarantee the presence and readiness of all vital components within the management cluster. For enhanced user experience, I should receive timely feedback via Custom Resources, apprising me of the readiness status of any specific release version of the cluster stack.
+
+### User Story 2: Intuitive Updates to Clusters via Cluster-API Native Mechanisms
+
+As a user, I expect a seamless and intuitive method to update my Cluster Stack. The ideal process should enable me to merely modify the reference point within the `ClusterClass` of the Cluster object. This is in line with the established protocols of Cluster-API's `ClusterClass`. Once the reference has been adjusted, the cluster-stack-operator should autonomously manage all subsequent steps, ensuring the successful update of the cluster addons within the appropriate workload cluster.
+
+### User Story 3: Effortless Access to the Latest Cluster Stack Versions via Github Integration
+
+I expect the cluster-stack-operator to be endowed with the ability to continuously poll Github releases to identify any new releases that might be available. As a user who values automation, when I activate the "auto-subscribe" functionality, I anticipate the operator to not just detect the latest release but also to execute all requisite steps to make this release fully operational and ready for my use.
+
+### User Story 4: Facilitated Use of Specific, Legacy Versions of Cluster Stacks
+
+As a user with diverse requirements, there might be instances where I might require a specific older version of a cluster stack. For such scenarios, I should be equipped to delineate specific release versions within the objects I deploy. This precision ensures that my desired versions remain accessible for my use, allowing for flexibility and backward compatibility.
+
+### User Story 5: Personalized Configuration of Provider-specific Parameters, such as Node Images
+
+As a discerning user, I wish to define provider-specific configurations to suit my unique needs. Within the framework provided by the API of the respective provider-specific operator, I should have the freedom to stipulate details like the exact node images I prefer. Such control is instrumental in optimizing resources, ensuring I can steer clear of redundant node images, and potentially avoid tasks like unnecessary uploads.
+
+### User Story 6: Local Testing and Building of New Cluster Stacks for Providers
+
+Recognizing the pivotal role of providers, it's essential that they possess the capability to develop and validate their cluster stacks locally. The cluster-stack-operator should be intelligent enough to discern the source of release artifacts. If these are furnished locally, the operator should abstain from downloading them from external sources like Github. Such a feature is paramount for providers—it ensures they can construct, test, and validate their cluster stacks in a local environment, guaranteeing robustness and operational excellence.
+
+## Background
+
+### Understanding the Three Core Components of a Cluster Stack
+
+In the realm of cluster management and orchestration, the cluster stack is designed to seamlessly integrate and operate various functionalities. There are three pivotal components of a cluster stack:
+
+
+There are three components of a cluster stack:
+
+1. **Cluster Addons:** Cluster addons are a set of essential extensions and tools, including but not limited to the Cluster Network Interface (CNI), Container Storage Interface (CSI), and Cloud Controller Manager (CCM). These addons play a pivotal role in enhancing the functionality and manageability of each workload cluster initiated by the user. It's paramount to ensure that these addons are applied to every single workload cluster upon initiation, thus enabling seamless communication and operational consistency.
+
+2. **Cluster Class:** The `ClusterClass` object acts as a backbone for the Cluster-API. This object simplifies the user's interaction with the Cluster-API by providing a standardized template. Within the cluster stack, the `ClusterClass` object is accompanied by various other Cluster-API objects that are indispensable for leveraging the full potential of the `ClusterClass`. For efficient cluster management and operation, these objects must be duly applied within the management cluster. This step ensures that users experience a standardized and streamlined cluster setup and management process.
+
+3. **Node Images:** Providing users with node images is not a one-size-fits-all proposition. These images can come in diverse formats tailored to specific needs and scenarios. What’s noteworthy is that these node images are meticulously released and tested in conjunction with the above-mentioned components of the cluster stack. This rigorous testing ensures compatibility, performance optimization, and a harmonized working environment between the node images and other stack components.
+
+In essence, the harmony and efficient interplay between these three components are what make the cluster stack a robust and reliable tool in clustered computing environments. Proper understanding and implementation of these components can significantly streamline operations and enhance overall cluster performance.
+
+### Feature Spotlight: The ClusterClass in Cluster-API
+
+Within the ecosystem of Cluster-API, a significant feature known as `ClusterClass` is prominently highlighted. This feature facilitates users in applying Cluster objects. When users engage with this feature, they can reference the Cluster objects directly from the Cluster-API, pointing specifically to the `ClusterClass` object. This streamlines the process, enabling a more efficient and integrated way to manage cluster objects.
+
+### A Comprehensive Guide: Crafting Node Images Across Various Providers
+
+Creating node images tailored for specific requirements becomes pivotal when deploying across different providers. There are primarily three distinct methods to achieve this. Let's delve deeper into each of them:
+
+1. **URL on a Remote Endpoint:** Some providers are equipped with the capability to deploy a Node Image directly from a URL. In such scenarios, the Node Image that is part and parcel of the cluster stack (especially located in the cluster class) would be designated via a URL. This URL acts as a pointer to a pre-built image that's hosted and available on a remote platform. The most striking advantage here is the convenience it offers. The operator essentially has no further tasks. Once the URL is provided, the rest of the process is managed by the system itself.
+
+2. **Utilizing Artifacts:** Certain providers are adept at supporting artifacts. When this is the case, the Node Image is translated into a tangible artifact. A common example is a qcow2 file. Post creation, this artifact is then uploaded onto the provider’s platform. When it comes to node provisioning within the cluster stack, this artifact becomes the reference point. However, this method demands a tad more involvement from the operator. They have to ensure that the artifact is accurately created, uploaded, and then appropriately referenced.
+
+3. **Relying on Build Instructions:** There are situations where providers might neither support the direct URL deployment nor the artifact-based provisioning. It's in such instances that the build instructions housed within a repository become invaluable. These instructions act as a detailed roadmap to build the Node Image. They meticulously outline all the required steps, configurations, tools, and best practices. For the operator, these instructions are paramount. They serve as the guidebook, ensuring that the operator can provide users with node images that adhere to the highest standards and meet specific needs.
+
+In summary, while the first method offers sheer convenience and a hands-off approach for the operator, the subsequent two methods do require a level of engagement. However, all three methods ensure that the end-users receive top-notch node images, tailored to their specific requirements.
+
+## Risks and mitigations
+
+### Challenges in Implementing Cluster Stack Operator Provider Integration
+
+#### Overview:
+
+For providers intending to make the cluster-stack-operator accessible to their clientele, they are tasked with the responsibility of implementing the provider-specific operator. This endeavor, while substantial, is not insurmountable. Let's explore the intricacies and safety measures associated with it.
+
+#### In-depth:
+1. **Prerequisite Experience for Providers:**
+    - Only those providers that have previously developed a Cluster API Provider Integration are mandated to craft the Cluster Stack Operator Provider Integration.
+    - This prerequisite acts as a quality assurance step. By ensuring that only those providers with prior experience in similar integrations take on this responsibility, we can enhance the probability of successful implementations.
+    - Providers utilizing a universally accepted framework like OpenStack are spared from this task. Instead, they can capitalize on shared efforts, collaborating with other CSPs that also employ OpenStack or similar technologies.
+2. **Exemptions Based on Provider-Specific Work:**
+    - In instances where there's no need for provider-specific interventions, like the building or uploading of node images by users, there's no necessity to craft the Cluster Stack Operator Provider Integration.
+    - Clarity is paramount here. Users must be well-informed that they can configure their resources in a way that aligns with these exemptions.
+
+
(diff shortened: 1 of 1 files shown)


jschoone:
Hi @SovereignCloudStack/dnation, @mxmxchere, @DEiselt, @NotTheEvilOne.
With this draft we have the first design document for the Cluster Stack Operator which will be a fundamental part of the future reference implementation.
Could you have a look on that?
Feel free to add comments/questions directly to the document here.
I'd like to set up a separate meeting to discuss this

There is also a draft for an [ADR on testing the cluster stacks](https://github.com/SovereignCloudStack/standards/pull/321) which can be treated in the same way.

Of course, if you have the capacity I would love it if everyone was there, but since those are two walls of text I don't expect everyone of you to read this, at least one person per tender would be great for now.

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+
+In a bid to offer users an intuitive experience and seamless transition, we've ensured that our object structure mirrors that of the Cluster-API. This alignment not only facilitates user understanding but also fosters a harmonious integration between the two closely related projects.
+
+However, there is a key distinction between our operator and the Cluster-API. Not all providers necessitate provider-specific work. Recognizing this, we've imbued our system with flexibility. There isn’t a compulsory need to construct a provider-specific operator. Instead, users have the autonomy to determine whether they wish to initiate such an operator based on their unique requirements. This approach affirms our commitment to offering a user-centric tool, tailored to meet diverse needs with precision.
+
+## User stories
```
mxmxchere, line 42: I am missing distinct user stories for the **CSP** offering the managed-k8s service and the **user** consuming managed-k8s service.

mxmxchere: There is User Story 1 with **SCS-user** is running the management cluster and **SCS-User** is consuming its self-hosted service. From my point of view it is crucial that we add and focus on a user story where offering and consuming of the managed-k8s are done by distinct players. (CSP and SCS-User)

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+
+#### In-depth:
+1. **Streamlining Information Flow:**
+    - As a solution, we aggregate all the pivotal information into two primary user-centric objects:
+        - The first is provider-specific and varies based on the chosen provider.
+        - The second enables users to detail the specifications of the cluster stack they desire.
+2. **Simplifying User Interactions:**
+    - Users need only interact with a singular object to specify all their requirements.
+    - This object also serves as the primary communication channel, updating users about the status of their cluster stack's release.
+    - The net result is a user-centric design where they engage with a minimal number of objects, while still benefiting from the operator's advanced features. The operator's intricate details remain in the background, ensuring users face minimal cognitive load.
```
mxmxchere, line 143: Very good, this is a clear improvement.

Regarding the missing user-story:

- From my point of view the **SCS-user** should only interact with exactly one object (per cluster)
- The **CSP** should manage the "provider-specific" object. It will feel very strange for an cloud-xyz customer to create cloud-xyz specific settings first before he/she can consume clusters

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+- apiVersion: clusterstack.x-k8s.io/v1alpha1
+    kind: ClusterStackRelease
+    name: myprovider-xyz-1-26-v6
+spec:
+  nodeImages:
+  - workeramd64
+  - controlplaneamd64
+```
+
+### Webhooks
```
mxmxchere, line 380: very-general and a waste of time for all readers to put that paragraph here

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+
+### Events
+
+Kubernetes events provide insight into the behavior of the components within the cluster. They record the changes or significant occurrences associated with objects in the Kubernetes ecosystem. We use them heavily in our controller and for our custom resources.  
+
+Here are some key points about Kubernetes events:
+
+1. **Purpose**: They give a chronological account of what's happening in the system. This can be crucial for debugging purposes, for understanding the lifecycle of resources, or for monitoring.
+2. **Scope**: Events can be associated with different types of objects. For instance, you might see events related to a Pod's lifecycle or events showing reasons why a service endpoint was not created.
+3. **Viewing Events**: The most common way to view events is by using the `kubectl` command-line tool:
```
mxmxchere, line 435: i am starting to feel fooled by now, have you used some LLM to generate this?

This is super unrelated and just taking my time

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+
+However, it's essential to be aware that our setup does have multiple external dependencies that need to be addressed to ensure comprehensive testing.
+
+To start with, we require a local kind cluster. This cluster functions as the foundation for our workload during local tests. By utilizing this kind cluster, we are endowed with the ability to emulate the process of applying cluster addons directly to a workload cluster.
+
+Moreover, for our testing regimen, we also have local releases of cluster stacks. These are particularly crucial for unit tests as they negate the need to rely on an external Git repository. In order to check that downloads work from e.g. Github, we add specific dedicated tests to make sure that the download works properly. 
+
+### Repo DevOps: Tools for Efficient Development
+
+#### kubebuilder
```
mxmxchere, line 483: yes kubebuilder might minimize boilerplate code, but this whole paragraph could have been left out...

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+
+Moving forward, the controller assumes an active role. Its first task is to enumerate all the `ClusterStackReleases`. This list is curated based on a specific criterion: every release must possess an owner reference, or, more aptly put, a controller reference set to the `ClusterStack`. This ensures that there's a structured hierarchy and organized linkage between different releases and their associated cluster stacks.
+
+The controller's role doesn't end with listing. It delves deeper, embarking on an analytical journey to ascertain the relevance and importance of each `ClusterStackRelease`. Several factors influence the controller's judgment:
```
mxmxchere, line 569: Seriously? Your controllers are embarking on analytical journeys? Great.

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+
+1. **Templating with Helm:** Our first step involves utilizing the helm binary to craft a template from the helm chart. Helm, as you might be aware, serves as a potent package manager for Kubernetes, and its capability to template charts is paramount for our objective.
+2. **Applying the Templated Objects:** Post templating, the objects are then methodically applied to the Kubernetes cluster using the client-go library, a standardized Go client for Kubernetes. As each object gets applied, we maintain a meticulous record in the resource status. This record provides clarity on the operational efficacy. If an object application is successful, its status is marked as synced. Conversely, if an object encounters a problem during its application, its status reads not synced. In such cases, we also provide insights into potential reasons for the failure, facilitating easier troubleshooting.
+
+Our approach resonates with methodologies adopted by renowned projects such as Argo-CD. We harness the power of both the helm binary and the Kubernetes Go client, ensuring a blend of robustness and flexibility.
```
mxmxchere, line 614: Is there a special reason not to use helm as a go-library?

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+
+The `ClusterStack` object embodies a set of core properties that facilitate Kubernetes cluster management and deployment. These properties include:
+
+1. **Provider:** This pertains to the platform facilitating the cluster. An example is OpenStack.
+2. **Kubernetes Version:** This delineates the version of Kubernetes being implemented.
+3. **Name:** A distinct identifier for the `ClusterStack`.
+
+A user engaging with the `ClusterStack` can command several operations, such as:
+
+- **Subscription to Git Repository:** Users have the autonomy to choose if they wish to auto-subscribe to updates of specific Git repositories. Upon this selection, the controller undertakes the responsibility of periodically interfacing with Git. Its aim? To vigilantly ascertain if fresh releases, pertinent to the said `ClusterStack`, have been dispatched. If it discerns any, an automatic download is initiated.
```
mxmxchere, line 187: Regarding the missing user-story:

This makes it very easy for **SCS-users** to create broken clusters that the **CSP** might have to fix. From my point of view it is better if the **CSP** manages a set of supported versions, or stacks that the **SCS-user** can choose from.  This enlarges the whole benefit of the cluster-stack being tested and the components being tweaked for each other.

mxmxchere (changes requested):
All in all this was very hard (and long) to read. As this is a technical document all marketing and bling-bling language could have been left out. Also some paragraphs about webhooks, events, kube-builder and declarative system management could just have been left out.

What is currently missing is a user-story for this being used in a role-separated setup (with **CSP** and **SCS-User** having interests and responsibilities for different resources). Regarding this setup we should also think about scaling, how many resources should one controller manage? Is there a mechanism planned form horizontal scaling (deploying multiple controllers and assigning resources to different controllers?)

Apart from that very good work and a great enhancement for the current state. This proposal solves two problems very well:

1. currently the user has to deploy multiple objects to get a running cluster
2.  the user has to install addons/extensions to get the working cluster

mxmxchere:
A real benefit to enhance my understanding of the new approach would be a picture showing the various objects with their contents (both cluster-apis as well as the new ones) and how they interact with each other. I think in most cases it is from top to bottom (cluster -> creating controlplane -> creating openstackcontrolplane and so forth). Also are there ideas on how to integrate with the planned central-api (https://github.com/SovereignCloudStack/issues/issues/364)? This would help to start thinking in the two roles "CSP" and "SCS-User"

jschoone:
> A real benefit to enhance my understanding of the new approach would be a picture showing the various objects with their contents (both cluster-apis as well as the new ones) and how they interact with each other. I think in most cases it is from top to bottom (cluster -> creating controlplane -> creating openstackcontrolplane and so forth). Also are there ideas on how to integrate with the planned central-api ([SovereignCloudStack/issues#364](https://github.com/SovereignCloudStack/issues/issues/364))? This would help to start thinking in the two roles "CSP" and "SCS-User"

:+1:  for the picture! We already have an issue for that:
https://github.com/SovereignCloudStack/issues/issues/333

fdobrovolny (commented):
Great work, guys.

I agree with @mxmxchere that this was not a light read, and I had to google some words, which is unnecessarily flowery.

Except for what I wrote in individual comments, some of which are my confusion, edit's to the formatting, dislike of a specific sentence, or possible improvements I see.

I would advocate adding a preface. I feel that this document walks through the whole architecture, and each point is addressed to every detail. I would like to see a preface where all the architecture is summarized with all the components in a few paragraphs so then when I'm reading the text, I can start connecting into context the individual parts.

Also, maybe connected to this, some kind of glossary would be nice. In a few words, summarize all the components in a list I can return to while reading.

The task [issues/#333](https://github.com/SovereignCloudStack/issues/issues/333) should be incorporated here as soon as possible.

I feel that the first reading is not enough. I will try to re-read this by the end of the week. Hopefully, I can get more profound insight.

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+
+4. **Safeguarding Node Images with a Dedicated Controller:**
+The `ProviderNodeImageRelease` object doesn't operate in isolation. It has its own guardian angel, a dedicated controller, ensuring that a specific node image is primed for the user. This is no simple task. The path to readiness often involves time-consuming operations, such as the meticulous uploading or intricate building of the node image.
+
+5. **Tackling the Challenge of Long-Running Operations:**
+These lengthy operations come with their own set of hurdles. Consider scenarios where the operator's container is abruptly halted by Kubernetes or manually terminated. The solution? A robust map securely stored in memory, vigilantly monitoring all concurrent operations (thankfully, the power of goroutines ensures parallel operations). Handling shutdowns with finesse, especially those triggered by SIGTERM, becomes paramount. The key lies in gracefully stopping all extended operations, potentially leveraging the canceling contexts in Go.
```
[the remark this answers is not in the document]
NotTheEvilOne: As a side note as for another controller I recently faced this exact issue again. There's no reliable way to execute shutdown actions within k8s even for pods under certain scenarios. Therefore the mentioned finesse must be added while handling with the data, not at shutdown time.

operations/kaas/cluster-stack-operator.md
janiskemper: I think there is a misunderstanding: The cluster stacks are all tested before they are released. This feature just makes sure that all users automatically get the latest released and tested cluster stacks. 
In case the management cluster is run by a CSP, it is less work for the CSP to update all the setups for the customers, as it is done automatically

operations/kaas/cluster-stack-operator.md
janiskemper: This can be done in case the CSP runs the management cluster, of course. This design document is general and is independent of who runs the management cluster. 
That the "user" has to do it only means that it is not done automatically and someone has to do it manually, whoever it is.

operations/kaas/cluster-stack-operator.md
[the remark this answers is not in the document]
janiskemper: agreed, the controller cannot do anything when it is killed. We then have to either handle some form of cleanup outside of the operator or when the operator re-starts.

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+## Background
+
+### Understanding the Three Core Components of a Cluster Stack
+
+In the realm of cluster management and orchestration, the cluster stack is designed to seamlessly integrate and operate various functionalities. There are three pivotal components of a cluster stack:
+
+
+There are three components of a cluster stack:
+
+1. **Cluster Addons:** Cluster addons are a set of essential extensions and tools, including but not limited to the Cluster Network Interface (CNI), Container Storage Interface (CSI), and Cloud Controller Manager (CCM). These addons play a pivotal role in enhancing the functionality and manageability of each workload cluster initiated by the user. It's paramount to ensure that these addons are applied to every single workload cluster upon initiation, thus enabling seamless communication and operational consistency.
```
NotTheEvilOne, line 77: As a more fundamental question: Maybe it makes sense to take one step back an review if at least part of the requirements should be implemented with tools already introduced to SCS, like flux2 [1]? This would make this document split into two parts as well as for flux maintenance will be more focused to setting up manifests.

It seems to me that partly the intended controller might reinvent the wheel.

[1] https://github.com/SovereignCloudStack/k8s-cluster-api-provider/pull/123

janiskemper: We take an operator approach and it is not clear how this approach would integrate with flux to me. 
The described ClusterAddons and their controller would make it possible to install the cluster addons without any manual work or configuration. 

Additionally, we have some values from Cluster API objects that we inject into the yaml config of the objects we apply. I wouldn't know how to do that with external software either. 

Can you maybe elaborate on your idea to integrate flux?

operations/kaas/cluster-stack-operator.md
garloff: We've had that discussion before indeed, and identified that both are scenarios we need to have in mind.
Situations, where someone external to the SCS-User manages the cluster (this might then be the CSP that also runs the infrastructure or a third-party which may be paid for this or a different department in the SCS-user's company). And situations where they are the same person. Even if it's the same person, that person has two different roles, so I think that the distinct look is the more general one.

operations/kaas/cluster-stack-operator.md
garloff: But let's keep things that providers need to configure (so it works on their infra) and things that consumers want to configure (so the cluster fits their needs) well separated please.

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+
+A key feature we utilize within the Cluster-API is the `ClusterClass`. The `ClusterClass` enables us to curate a comprehensive "cluster bundle". Within this bundle, we encompass all the vital components needed for a fully functional, production-ready cluster. This includes the kubeadm configurations and other crucial applications. Users can effortlessly create a custom cluster by simply referencing the desired `ClusterClass` object to the Cluster.
+
+In order to solve some issues that go beyond core CAPI components, we bundle application configuration, node images, and the `ClusterClasses` to formulate what we term as "cluster stacks".
+
+While the Cluster-API is adept at managing the lifecycle of clusters, there are aspects it doesn't cover. For instance, it doesn't oversee the lifecycle of the Cluster Stack components. For example, when a new Syself Cluster Stack is introduced, the corresponding `ClusterClass` object should be downloaded and implemented, ensuring it's readily accessible to users. Additionally, there are certain tasks that are provider-specific, such as building node images, that the Cluster-API doesn't cater to.
```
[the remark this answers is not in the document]
garloff: deployed? applied?

operations/kaas/cluster-stack-operator.md
janiskemper: What do you mean @garloff ? This operator could be used by anyone having access to cluster stacks. If it is a public cluster stack, then anyone can use it - no matter if they are CSPs or not. 
This design works for both cases.
And if you are not a CSP, you need to take care of the provider-specific things as well. 

Which might be, btw, something like the node images that you want to have. This is user-specific and the CSP can potentially leave that to the users (if they want).

operations/kaas/cluster-stack-operator.md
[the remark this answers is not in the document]
joshmue: Ideally, there should not be any long-running operations in controllers at all, right?

If an operation really is long-running, tracking progress using usual (mostly stateless) reconciliation loops (setting `status` to track preliminary states) should be the K8s native way to do it right? Maybe even splitting out some things into their own resource definitions with their own reconciliation loops?

That may indeed prove hard to do if some long-running synchronous call has to be executed - but if that is not something that is already foreseeable, IMHO this probably is not something to worry about beforehand.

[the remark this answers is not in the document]
janiskemper: Agreed that there shouldn't be long-running operations at all. But here we cannot avoid it, as building or uploading images might be long-running. In this case we use goroutines that work in parallel and that don't block the reconciliation loop. 

The only other alternative is starting separate containers via Kubernetes jobs, but that would be quite messy as well.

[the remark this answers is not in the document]
batistein: Creating a Kubernetes job might appear simple initially, but it can lead to more overhead than anticipated. Here's why:

At first glance, creating a job in Kubernetes seems efficient because it manages the lifecycle of your task. But, there are challenges:

1. Monitoring: If the job or the pod fails, how would you know? For reliability, you'd need a monitoring solution to check the job's health.
1. Error Recovery: If a job gets stuck, you would need logic to restart it.
1. Data Cleanup: If your system identifies images by their names, you might end up with unused or "orphaned" images. A garbage collector would be needed to clean them up.
1. State Information: You would also need a way to regularly check the status or progress of the pod. This could involve accessing the pod directly or having the job send status updates.
1. User Visibility: To make the system user-friendly, you'd need a way to show users the progress or status of their tasks.

Considering these complexities, it might be simpler to directly run a pod and manage its lifecycle with a controller. However, this introduces a testing challenge. Unit tests may not be sufficient, and tools like envtest can't handle this scenario. Therefore, you'd rely on end-to-end (e2e) tests. These tests are slower, tougher to maintain, and can be time-consuming when identifying issues.

You might then wonder: "Why run this as a separate pod at all?" Is it to estimate resource needs or to integrate more smoothly with other systems?

An alternative is using Go routines. Here are their benefits:

1. Easy Testing: You can test the logic easily, and native tools work out-of-the-box.
1. Simplified Management: The lifecycle is managed directly in the code. There's no need to interact with the Kubernetes API or set up complex RBAC.
1. Real-Time Information: Getting updates about the ongoing task is straightforward.
1. Resource Management: Go routines can be queued, helping predict resource usage. Since in our use case one controller won't spawn hundreds of goroutines simultaneously, it's efficient and avoids unnecessary overhead.

Conclusion: If you're aiming for a quick fire and forget method, a Kubernetes job is suitable. However, for building robust systems, alternative methods like Go routines are more straightforward and reliable. They offer precise control over long-running tasks, ensure easier maintainability, and reduce the intricacies of testing. This leads to better code quality in less time. Of course, using an approach with creating pods is entirely feasible. If we were operating on a larger scale, this would indeed be the preferred solution. However, it comes with all the previously mentioned challenges, adding unnecessary overhead for our current needs. As the saying goes, usually the simplest solution is the best! ;)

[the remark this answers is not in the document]
joshmue: > But, there are challenges

As always :D 

> As the saying goes, usually the simplest solution is the best! ;)

I can also agree to this statement, while I still think that it actually favours a `Job` approach. Mirroring your points:

> Monitoring: If the job or the pod fails, how would you know? For reliability, you'd need a monitoring solution to check the job's health.

A monitoring solution is not strictly required, as [`JobStatus v1`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#jobstatus-v1-batch) includes `conditions` (`Failed`/`Complete`).

> Error Recovery: If a job gets stuck, you would need logic to restart it.

For clean exits, this is actually builtin (and the primary purpose of a Job compared to a one-off Pod): "A Job creates one or more Pods and will continue to retry execution of the Pods until a specified number of them successfully terminate." (https://kubernetes.io/docs/concepts/workloads/controllers/job/ CC BY 4.0).
For cases, where a non-zero-error exit is not done by the process on its own (-> being "stuck"), I guess using `livenessProbes` would be the way to go, forcing a non-zero-error return.

> Data Cleanup: If your system identifies images by their names, you might end up with unused or "orphaned" images. A garbage collector would be needed to clean them up.

Yes, cleaning up is required, indeed (even though some cleaning up will also be required when implementing in-controller-process-jobs, if leakage should be prevented). If not done in the controller itself, the builtin TTL for Jobs can be used.

> State Information: You would also need a way to regularly check the status or progress of the pod. This could involve accessing the pod directly or having the job send status updates.

Querying [`JobStatus v1`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#jobstatus-v1-batch) should be sufficient, I guess.

> User Visibility: To make the system user-friendly, you'd need a way to show users the progress or status of their tasks.

That indeed would not be something that is builtin into Kubernetes already, but I am also unsure how something like progress reporting (or progress bars) would be implemented when using in-controller-process-jobs. Would the controller update `status` of Cluster Stack Releases, notifying about stages of image building and pushing?

---

Regarding the use of Go routines for implementing in-controller-process-jobs: I like Go routines very much as well and have a fair share of experience using them, but here are a few disadvantages to this approach:

1. One would still have to implement a non-trivial lifecycle system - which is easier to do in Go than in most other languages, indeed. But it will still be a non-trivial problem implemented in a stateful process that may be terminated during e.g. node reboots. So, if any bug appears...
2. ...identifying a problem is much more complicated for anyone who does not have a Go debugger (or dev environment at least) at hand in e. g. production. When having almost all state inside of K8s API objects, problems are found much easier - for anyone familiar with K8s, but not necessarily with Go, Go routines and the to-be-built lifecycle system on top.

So, at least this point:

> Simplified Management: The lifecycle is managed directly in the code. There's no need to interact with the Kubernetes API or set up complex RBAC.

...has a large flip side. I think that interacting with the K8s API is actually the most natural fit for any controller/operator, and that RBAC actually is not too complex most of the time.

Privilege escalation via Job/Pod creation could be a real tradeoff, though.

operations/kaas/cluster-stack-operator.md
```diff
@@ -0,0 +1,692 @@
...
+9. **Event Exporters**: There are tools and systems like Prometheus and Eventrouter that can be set up to scrape or receive these events and use them for monitoring, alerting, or logging purposes.
+
+In practice, Kubernetes events are a vital tool for developers and operators to understand what's going on inside their cluster, diagnose issues, and monitor the health and behavior of their applications and resources.
+
+We believe that events are, together with the conditions in the status of the event, the two sources that users should know in order to understand the system. Events might show some form of development of a certain incident over time, e.g. that the user understands that something has been going wrong 10 minutes ago and that is why the system is still stabilizing, but that the problem is already gone. 
+
+The conditions, on the other hand, should always show the current information that is available to describe the state of a resource.
+
+### Controllers that react immediately on specific etcd events
```
joshmue, line 454: `etcd events` -> `API server updates`?

(`etcd` should be abstracted away, `events` may be ambiguous with Kubernetes Event Resources)

operations/kaas/cluster-stack-operator.md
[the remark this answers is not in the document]
joshmue: > Privilege escalation via Job/Pod creation could be a real tradeoff, though.

I guess that this strongly relates also to @fdobrovolny's [remark](https://github.com/SovereignCloudStack/docs/pull/50/files#r1301932822) about the image building process being relatively unspecified.

Building images inside of a container is in my experience usually a rather complicated task, when it comes to security and isolation. The controller may need more privileges on the host system or may schedule e. g. dedicated one-off kaniko Pods. That would be still subject of research/decision making, I guess?

PR #55, opened 2023-08-29
Add source code of docs-page/main

maxwolfs:
This PR is migrating the source code of our [docs-page](https://docs.scs.community) from https://github.com/SovereignCloudStack/docs-page to this https://github.com/SovereignCloudStack/docs repository. 

Aim of this is to have one single source of truth for the content docs files of the @SovereignCloudStack Organisation. This move is part of implementing this [standard](https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0002-v2-standards-docs-org.md)

This will be a `merge-commit` to preserve the history and must thus not be squashed.


garloff (approved):
Obviously, I can't re-review all the imported changes from the docs-page repo.
What I'm trying to judge is that the merge has been done correctly and at least the obvious fallout to have been sorted.
This appears to be the case, so I approve.
(Sidenote: We followed: https://stackoverflow.com/questions/1425892/how-do-you-merge-two-git-repositories)
Reminder: Don't *squash* merge this -- we want to preserve history.

garloff:
github does not seem to allow a regular merge commit with the settings that we have chosen.
We need to figure out what settings need to change ...

PR #69, opened 2023-08-30
Fix standard links

### LINKED ISSUE #67: 404: https://docs.scs.community/standards/scs-0102-v1-image-metadata/
Our public documentation at https://docs.cleura.cloud/ links to the SCS image metadata standard, which until yesterday (2023-08-22) used to be available at <https://docs.scs.community/standards/scs-0102-v1-image-metadata/>. As of today, that link has apparently turned into a 404.

This being classified as a "Stable" standard per <https://github.com/SovereignCloudStack/standards/blob/main/Standards/scs-0102-v1-image-metadata.md>, I am guessing the corresponding documentation URL is also meant to be stable, thus I am suspecting that the removal was unintentional — though I am also not quite sure how SovereignCloudStack/docs-page#91, which was the only PR that landed on the day in question, could have caused this.

Could the original link be reinstated, or if that is not possible, could a redirect from the pre-existing to any updated link be put in place?

See also: https://github.com/citynetwork/docs/issues/228

maxwolfs:
fixes #67


garloff (approved):
Thanks for fixing this!

PR #70, opened 2023-08-31
Add redirect rule for old community pages url

maxwolfs:


artificial-intelligence (approved):
LGTM but I'm not deeply familiar with docusaurus code.

PR #71, opened 2023-09-01
Clean up after migration

maxwolfs:
* some sorting of markdown files into the new folder structure where they fit
* fixing the place of where `status page api` is being put into via the workflow
* open question: where to put the LBaaS document? (distinction between scs dev docs and scs operator/integrator docs still needs improvement)

Preview Deployment: https://docs-staging.scs.community/docs


fdobrovolny (approved):
LGTM

PR #87, opened 2023-10-09
Add 5G Campusnetz as usecase / deplyoment example

maxwolfs:


berendt:
After restructuring the OSISM docs for better integrability: does it makes sense to add this here: https://docs.scs.community/docs/iaas/guides/deploy-guide/examples/

berendt:
How to proceed here? Would be nice to be able to reference this deployment example.

rholling:
Yes, I am on the same page. We're actually moving our docs into a new wiki. 
The next step would be to check (ourside), which docs we're able to share and: how to share/integrate in github.
I will contact Max within the next days regarding further proceedings...

garloff:
How do we get this finalized, so we can merge it?

garloff:
I've just merged it into the staging branch for better review. (Give it 10mins to build and show up on https://docs-staging.scs.community/).
I'm not sure the autogenerated index building does work as wanted, so I'd need to have at least a look at the sideBar to see if IaaS/Deployment Examples shows four documents ... (artcodix, Testbed, CiaB, and ths 5G Campusnet). If not, we'll need to explicitly list them. (I would then also add a pointer to the turnkey HW Lab.)

garloff:
The autogenerated index seems to work.
Good to see the 5G Campus Net docs here!
Reading through it, it does however raise the expectation that there is a lot more to come.
@rholling - I want to follow your guidance here:
- Should we merge as is to have a first set of docs and hope to get more in the future?
- Or should we expect you to push a lot more in the next days and wait for that before merging?

rholling:
@garloff Exactly: please don't merge yet, as our source systems and the underlying hardware are currently moving. I will push more documents after the move.

PR #98, opened 2023-11-02
Add Container Introduction

### LINKED ISSUE #95: Prepare Container layer introduction
Result from SIG Documentation

* We need an introduction text for container layer answering the following questions on: https://docs.scs.community/docs/category/container-layer
    * Who is the target group? 
    * What is it not?
    * Prerequisites and Requirements (knowledge, software, hardware)
    * Features, Capabilities and Limitations (k8s cluster api provider only runs on openstack)
        * how does it interact with the iaas layer?
    * current plans on implementations, roadmap and current state on clusterstacks 
        * is it worth it to implement with the outlook on clusterstacks?
        * comparison table of cluster-api-provider and cluster-stacks
        * alpha label might make sense for cluster-stacks to indicate that it's in development
    * Success Stories and Deployment Examples

https://input.scs.community/container-layer-intro-on-docs-page?edit

maxwolfs:
written by @fdobrovolny – Thanks!

As discussed in the SIG Docs, adding some more information of the upcoming cluster-stacks would be great.

closes #95

### DIFF
--- a/docs/03-container/index.md
+++ b/docs/03-container/index.md
@@ -1,0 +1,43 @@
+# Container Layer Introduction
+
+The container layer within the Sovereign Cloud Stack (SCS) offers a robust solution for managing container workloads on a Kubernetes infrastructure. It facilitates the on-demand creation and scaling of Kubernetes clusters, catering to various needs across development, testing, deployment, and operation of services and applications. While the container layer is versatile for a range of use cases, the most common ones include:
+
+- Self-Service: Here, entities can empower their development teams with the autonomy to spawn and manage their clusters on demand, fostering a more agile and responsive development environment.
+- KaaS (Kubernetes as a Service): In this model, an SCS Cloud provider can offer their customers a managed Kubernetes service, abstracting much of the underlying operational complexity from their customers.
+
+## Target groups
+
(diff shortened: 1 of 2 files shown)


fdobrovolny:
@SovereignCloudStack/syself could someone from you guys please add some section and information about clusterstacks? Maybe some comparison of the two?

berendt:
Does it makes sense to add a "Vision" section as well? Not sure if the @fdobrovolny  can write this, more like an idea for this kind of document.

fdobrovolny:
@berendt 
> Does it makes sense to add a "Vision" section as well? Not sure if the @fdobrovolny can write this, more like an idea for this kind of document.

Could you explain more about how you mean it?
Do you mean vision as a kind of road map or more like an ultimate end-goal of a perfect solution, or more like philosophical what values should it have and what should it enable its users to do?

berendt:
The idea would be to describe where we see the container layer in 18 months. But let's maybe better take that separately and add to it later.

janiskemper:
I can add something @fdobrovolny but I'm not sure how you mean it. Do you want to include that information also in this PR or another one that I can work  on after this is merged?

fdobrovolny (approved):

fdobrovolny:
@janiskemper I discussed this with @maxwolfs, and we will merge this. Could you please make a new PR extending this?

janiskemper:
Here is the issue: https://github.com/SovereignCloudStack/docs/issues/99

PR #126, opened 2023-12-14
Reference public calendar instead of duplicating meeting info

garloff:
This should give an overview over the structure of meetings, but the precise times should not be duplicated here. Rather refer to the public calendar. Being at it, also clarify a few details on Nextcloud etc.


fkr (approved):
thanks  @garloff - very good.

PR #132, opened 2024-01-17
Bugfix: standard version being showed both stable and deprecated

mbuechse:
resolves [standards/#446](https://github.com/SovereignCloudStack/standards/issues/446)


maxwolfs (approved):
LGTM

PR #133, opened 2024-01-17
Update p500924-sig-monitoring1 'needed until' date

matofeder:
This project now hosts the productive monitoring cluster, hence we need to keep it alive


frosty-geek (approved):
LGTM

PR #148, opened 2024-02-27
fix two broken links in the faq

artificial-intelligence:
Signed-off-by: Sven Kieske <kieske@osism.tech>

### DIFF
--- a/docs/08-faq/index.mdx
+++ b/docs/08-faq/index.mdx
@@ -46,7 +46,7 @@ IaaS stands for Infrastructure as a Service, providing virtualized computing res
 
 ## How is the IaaS layer structured?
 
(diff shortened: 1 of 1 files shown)


artificial-intelligence:
actually there are many more instances where doc.scs.community is linked but these all result locally in `NS_ERROR_UNKNOWN_HOST` and this seems to be a global problem at the time of testing:

https://dnschecker.org/#A/doc.scs.community

is this intentional or a misconfigured DNS?

artificial-intelligence:
converted to a draft until it is clear that it's the code that needs fixing. also this fix is incomplete as mentioned above.

maxwolfs (approved):

PR #151, opened 2024-02-29
Extend playgrounds

jschoone:
Extend the playgrounds so we can not only use them at the KaaS-Hackathon in Bratislava (March 5+6) but also at the SCS-Hackathon in Arnsberg (April 16)


frosty-geek (approved):
LGTM

PR #154, opened 2024-03-04
Add general guidelines for writing test scripts regarding RFC 2119 Ke…

cah-patrickthiem:
This PR adds a document for general guidelines in regard of writing test scripts, especially on how to use RFC2119 keywords.

Resolves: SovereignCloudStack/standards#447


martinmo:
@cah-patrickthiem is the text ready for review from your point of view or are you still working on it? If you're finished feel free to request a review from me :-)

I already have a few remarks/suggestions:

- Your change appears to be very out of sync with the current state in `main`: `contributor-docs/` should be used instead of `dev-docs/`.
- Did you discuss adding the numbering scheme to the subfolders? I am asking because this might break existing links.
- To remain consistent with the other files in this repo, `RFC2119-keyword-test-guide.md` should be completely lower case (slug format).

mbuechse (commented):
Not a real review, just a remark and a request: once these guidelines are merged, please add a link to <https://github.com/SovereignCloudStack/standards/blob/main/.github/ISSUE_TEMPLATE/stabilize-standard.md?plain=1#L15>

cah-patrickthiem:
> @cah-patrickthiem is the text ready for review from your point of view or are you still working on it? If you're finished feel free to request a review from me :-)
> 
> I already have a few remarks/suggestions:
> 
> * Your change appears to be very out of sync with the current state in `main`: `contributor-docs/` should be used instead of `dev-docs/`.
> * Did you discuss adding the numbering scheme to the subfolders? I am asking because this might break existing links.
> * To remain consistent with the other files in this repo, `RFC2119-keyword-test-guide.md` should be completely lower case (slug format).

You are completely right here. The branch is out of sync. Also, the subfolders were just a quick catch of me, I understand that this needs to be discussed, but right now I also think this is not necessary.
I will change and fix the things you suggested since I wanted to tackle this anyway.

cah-patrickthiem:
Pushed a !fixup to address several inconsistensies. @martinmo the text itself is ready for review, feel free to check it. Also, if you want to add more content regarding tests, feel free to do so.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,57 @@
...
+* **MAY**: This is more flexible and could be used to provide additional
+information or best practices. Their impact on compliance is not as direct as "MUST" or
+"SHOULD" categories. These can be marked as blue.
+
+## 3. Channels for Output in Test Scripts
+
+* Align the RFC2119 keywords with specific channels in test scripts:
+  * "MUST" and "MUST NOT" should lead to errors
+  * "SHOULD" and "SHOULD NOT" to warnings
+  * "MAY" to notes or hints.
```
mbuechse, line 43: We should be more specific about these channels.
I would assume the relevant output to happen on `stderr`, and the channel is indicated by the first word on the line, so the line would start with `ERROR` (ideally all uppercase), for instance, followed by a colon or a whitespace character.
the channel for "SHOULD" would be `WARNING`,
the channel for "MAY" would be `INFO`,
and there would be a fourth channel called `CRITICAL` for circumstances where the test cannot be performed.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,57 @@
...
+
+## 4. Compliance and Test Passing Criteria
+
+* Emphasize that for a test to pass, it must meet all the "MUST" requirements without
```
mbuechse, line 47: The beginning reads like a prompt for someone who should actually write the text -- did you by any change forget to run it through an LLM?

I would write it like so:

> A test is considered to pass if and only if it doesn't produce any messages on the ERROR nor the CRITICAL channel.

Or, maybe even like so:

> A test can have one of three results:
>
>  - fail: if it has at least one message on ERROR
>  - DNF (did not finish): if it has no message on ERROR, but at least one on CRITICAL
>  - pass: otherwise
>
> The messages on the other channels serve informational purposes, for instance, how well the test subject might perform with future versions of the standard.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,57 @@
...
+errors. The presence of warnings or informational items from "SHOULD" and "MAY" 
+categories does not directly impact the pass/fail status but is important for overall 
+quality and compatibility with future standards.
+
+## 5. Conclusion and Continuous Improvement
+
+* Summarize the importance of adhering to these guidelines for maintaining compliance 
+and adaptability within the SCS framework. Highlight the need for ongoing updates and 
+revisions to the document to align with evolving standards and best practices in cloud 
+infrastructure and services.
```
mbuechse, line 57: This again reads like a prompt.

mbuechse:
After the remarks I just made, I have a suspicion that these docs may actually become a standard one day. But it's fine to start out with docs.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
mbuechse: The final part cannot be stressed enough: if the test cannot be performed for whatever reason, a message MUST be emitted on CRITICAL, and this means: definitely catch all exceptions!

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
cah-patrickthiem: Agreed, for now I excluded this section since for me it does not really makes sense here anyway.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
cah-patrickthiem: Agreed, for now I reworded the first sentence.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,156 @@
...
+
+* Firewall Setup: The firewall **SHOULD** be configured to allow only necessary ports,
+such as 80 (HTTP) and 443 (HTTPS). All unnecessary ports **SHOULD NOT** be open.
+
+* Backup Plan: Regular backups of the web server data **MAY** be performed, and it is
+recommended that backup integrity checks are conducted to ensure recoverability.
+
+#### Compliance
+
+Failure to comply with these requirements results in the web server being
```
martinmo, line 100:
```suggestion
Failure to comply with these requirements results in the web server installation being
```

As I understand it, the example standard is about the installation (and configuration) of the webserver and its environment and not the webserver itself.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,156 @@
...
+## Introduction
+
+The process of test creation and standards definition in Sovereign Cloud Stack (SCS)
+involves developing and validating tests that align with the specific requirements and
+standards set by SCS. This entails a comprehensive understanding of the SCS framework,
+which is built on open-source components and aims to provide interoperable and sovereign
+cloud services. In creating tests for SCS, it's essential to consider the various aspects
+of cloud technology, including infrastructure, services, and the integration of these
+components. The tests should be designed to ensure that the SCS standards are met, which
+includes adhering to specified requirement levels as indicated by [RFC2119 keywords](https://datatracker.ietf.org/doc/html/rfc2119).
```
martinmo, line 17: The introduction is good but also very generic. In my opinion, it does not yet answer important questions: Why is it needed/why should I read it? Who should read it (or to whom is it important)?

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,157 @@
...
+
+        # success log
+        logger.debug("Process completed successfully")
+
+    except ValueError as e:
+        # debug log before raising a critical error
+        logger.debug("Encountered a critical error about to log", exc_info=True)
+        logger.critical(f"Process failed: {e}", exc_info=True)
+
+```
```
martinmo, line 155: IMHO somewhere in this document it should be documented that the exit code of the conformance test script must be non-zero if we have ERROR or CRITICAL messages.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,157 @@
...
+- **MAY** (same as **OPTIONAL**)
+
+to define requirements for SCS-compatible IaaS and KaaS resources. These standards are crucial for ensuring interoperability and sovereignty in cloud services, as they are built on open-source components such as Kubernetes.
+
+To ensure readability and comprehensibility, only the main keywords **MUST**, **MUST NOT**, **SHOULD**, **SHOULD NOT** and **MAY** are referred to below.
+
+## 2. Interpreting RFC2119 Keywords in the Context of SCS
+
+- **MUST and MUST NOT**: These are binary and straightforward to test. Compliance or
+  non-compliance directly affects the standard meeting. For visualization, fulfilled
```
martinmo, line 36:
```suggestion
  non-compliance directly affects whether the standard is satisfied. For visualization, fulfilled
```

(To me, *standard meeting* sounds like something were people meet.)

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,157 @@
...
+- **MAY**: This is more flexible and could be used to provide additional
+  information or best practices. Their impact on compliance is not as direct as **MUST** or
+  **SHOULD** categories. These can be marked as blue.
+
+## 3. Channels for Output in Test Scripts
+
+- Align the RFC2119 keywords with specific channels in test scripts:
+  - **MUST** and **MUST NOT** lead to the ERROR channel for failed tests
+  - **SHOULD** and **SHOULD NOT** lead to the WARNING channel, the test is still passed
+  - **MAY** to leads to the INFO channel, the test is still passed
```
martinmo, line 51: The output channels seem to come out of "nowhere", have you considered introducing them briefly beforehand? And also why they are used (this might be good to know).

Something along the lines of 

> In test scripts, we use *channels* to convey information of different importance to the user. These channels are based on the ubiquitous and de-facto standard logging levels of common logging libraries in scripting and programming languages, in particular Python: DEBUG, INFO, WARNING, ERROR, and CRITICAL.

Furthermore, because the first enumeration level just has a single item ("- Align the RFC2119 keywords ..."), I think it should be "flattened".

martinmo (changes requested):
I reviewed it and left some suggestions for some minor issues I have found. Overall this looks really promising and we should get this merged soon :+1:

martinmo:
One more thing, because I just read @mbuechse's comment https://github.com/SovereignCloudStack/docs/pull/154#discussion_r1525437824 again.

We should definitely discuss whether to specify in the document that all *channels* output should go to `stderr` (for Python logging, this is the default) and what kind of format we expect (`<CHANNEL>: <MESSAGE>`).

I'm in favor of putting it into this document.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
cah-patrickthiem: Here I disagree, if the installation is not followed strictly and therefore not compliant, the webserver itself will be not compliant with these rules.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
cah-patrickthiem: I added this as well.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
cah-patrickthiem: changed

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
cah-patrickthiem: Added the explanation. 
Changed the enumeration level.

cah-patrickthiem:
> One more thing, because I just read @mbuechse's comment [#154 (comment)](https://github.com/SovereignCloudStack/docs/pull/154#discussion_r1525437824) again.
> 
> We should definitely discuss whether to specify in the document that all _channels_ output should go to `stderr` (for Python logging, this is the default) and what kind of format we expect (`<CHANNEL>: <MESSAGE>`).
> 
> I'm in favor of putting it into this document.

Added this as well.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,173 @@
...
+
+That means a test can have one of three results:
+
+- fail: if it has at least one message on ERROR
+- DNF (did not finish): if it has no message on ERROR, but at least one on CRITICAL
+- pass: otherwise
+
+In addition, a test **MUST** exit with a non-zero exit code (e.g., via `sys.exit(…)`) if there are any ERROR or CRITICAL messages, thus signaling a failure to meet a standard.
+
+It is crucial to determine and explicitly state within this document the expected conventions for logging output, specifically regarding the redirection of all channel outputs to standard error (**stderr**), which aligns with Python's default logging behavior. The format which **MUST** be used is **CHANNEL(MESSAGE)**, where **CHANNEL** represents the log level (e.g., DEBUG, INFO, WARNING, ERROR, CRITICAL) and **MESSAGE** encapsulates the actual log message.
```
martinmo, line 70: > It is crucial to determine and explicitly state within this document the expected conventions for logging output

This sounds like a prompt for the reader to determine the conventions and should be shortened to something like "the scripts are expected to ...".

Instead of `CHANNEL(MESSAGE)` the actual format we use is `CHANNEL: MESSAGE`. Example:

    INFO: Checking cluster specified by context 'my-cluster' in /path/to/.kube/config.

martinmo (changes requested):

martinmo:
With regard to my comment:

> We should definitely discuss whether to specify in the document that all channels output should go to `stderr`

I double checked with the existing scripts in the standards repo. Some scripts explicitly set the logging stream to `sys.stdout` ([example](https://github.com/SovereignCloudStack/standards/blob/c95b917a4a5387ceb6d96dd1e1d973caf13b51bb/Tests/kaas/k8s-node-distribution/k8s-node-distribution-check.py#L56)). Also, the "umbrella script" `scs-compliance-check.py` [supports both `stdout` and `stderr`](https://github.com/SovereignCloudStack/standards/blob/c95b917a4a5387ceb6d96dd1e1d973caf13b51bb/Tests/scs-compliance-check.py#L182-L187).

Maybe we should leave the `stderr` part out and allow both. Otherwise, we need to adapt the affected scripts to conform to this new guideline :-) @mbuechse do you have a strong opinion here?

mbuechse:
> Maybe we should leave the `stderr` part out and allow both. Otherwise, we need to adapt the affected scripts to conform to this new guideline :-) @mbuechse do you have a strong opinion here?

I wouldn't say strong, but I think the bit with stdout is a mistake, and stderr should be a recommendation at the very least.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,173 @@
...
+  with a valid reason for non-fulfillment.
+- **MAY**: This is more flexible and could be used to provide additional
+  information or best practices. Their impact on compliance is not as direct as **MUST** or
+  **SHOULD** categories. These can be marked as blue.
+
+## 3. Channels for Output in Test Scripts
+
+In test scripts, different channels are used to convey information of different importance to the user. These channels are based on the ubiquitous and de-facto standard logging levels of common logging libraries in scripting and programming languages, in particular Python: DEBUG, INFO, WARNING, ERROR and CRITICAL.
+
+Alignement of the RFC2119 keywords with specific channels in test scripts:
```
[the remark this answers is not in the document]
cah-patrickthiem: done

cah-patrickthiem:
> > Maybe we should leave the `stderr` part out and allow both. Otherwise, we need to adapt the affected scripts to conform to this new guideline :-) @mbuechse do you have a strong opinion here?
> 
> I wouldn't say strong, but I think the bit with stdout is a mistake, and stderr should be a recommendation at the very least.

Fyi I talked to @martinmo and we came to the conclusion to make "stderr" a MUST for test scripts.

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
cah-patrickthiem: done

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,173 @@
...
+
+That means a test can have one of three results:
+
+- fail: if it has at least one message on ERROR
+- DNF (did not finish): if it has no message on ERROR, but at least one on CRITICAL
+- pass: otherwise
+
+In addition, a test **MUST** exit with a non-zero exit code (e.g., via `sys.exit(…)`) if there are any ERROR or CRITICAL messages, thus signaling a failure to meet a standard.
+
+It is crucial to determine and explicitly state within this document the expected conventions for logging output. In test scripts it is a **MUST** to redirect all channel outputs to standard error (**stderr**), which aligns with Python's default logging behavior. The format which **MUST** be used is **CHANNEL: MESSAGE**, where **CHANNEL** represents the log level (e.g., DEBUG, INFO, WARNING, ERROR, CRITICAL) and **MESSAGE** encapsulates the actual log message.
```
martinmo, line 70:
```suggestion
In test scripts it is a **MUST** to redirect all channel outputs to standard error (**stderr**), which aligns with Python's default logging behavior. The format which **MUST** be used is **CHANNEL: MESSAGE**, where **CHANNEL** represents the log level (e.g., DEBUG, INFO, WARNING, ERROR, CRITICAL) and **MESSAGE** encapsulates the actual log message.
```

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
```diff
@@ -0,0 +1,173 @@
...
+
+For instance, when logging an ERROR regarding a Kubernetes cluster, the format on the command line should be like this:
```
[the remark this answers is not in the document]
cah-patrickthiem: done

contributor-docs/operations/tests/rfc2119-keyword-test-guide.md
cah-patrickthiem: done

martinmo (approved):
LGTM!

Issue #147, opened 2024-02-27
Open Stack Health Monitor Guide


maxwolfs:
tba

garloff:
Fixed by merging #155.

PR #155, opened 2024-03-06
Add OS Health Mon setup guide.

garloff:
This should address #147, once complete.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 /docs/02-iaas/deployment-examples/testbed
 /docs/03-container/components
 /docs/04-operating-scs/components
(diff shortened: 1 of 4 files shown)


maxwolfs:
It's working now: https://docs-staging.scs.community/docs/operating-scs/guides/openstack-health-monitor/Debian12-Install

I have changed the target directory to macht the existing strucutre. But the second step that was missing was to add imported files to the sidebarsDocs.js file to appear in the side navigation.

maxwolfs (approved):

garloff:
Thanks a lot, @maxwolfs  !

PR #160, opened 2024-03-14
Adjust populateStds.js and populateCerts.js to renamed field

mbuechse:
(deprecated_at instead of obsoleted_at)

see https://github.com/SovereignCloudStack/standards/commit/947fa5e57a25266456ae315a718266f67c3d2da5


maxwolfs (approved):
LGTM

PR #164, opened 2024-03-18
Link to v2 test for PCO1,2,3. Add PCO4 and AOV.

garloff:
Also include direct links to dashboards with mycloud set correctly for the PCO cloud regions (they are all on the same grafana host)..


maxwolfs (approved):
lgtm

maxwolfs:
The Cert of https://apimon.services.regio.digital/public-dashboards/17cf094a47404398a5b8e35a4a3968d4 has expired.

garloff:
> The Cert of https://apimon.services.regio.digital/public-dashboards/17cf094a47404398a5b8e35a4a3968d4 has expired.

Yeah, I noticed the failed CI action and sent a notification to Christian last night. We should not wait for the fix though, IMVHO, as our PR is not causing the regression. (As the CI pipeline stops, we take a small risk that tests happening after the failed cert don't run and thus we don't see the outcome.)

PR #167, opened 2024-03-20
Create default-security-group-rules.md

josephineSei:
a guide for default security group rules

follow-up part of: https://github.com/SovereignCloudStack/standards/issues/521

### DIFF
--- a/docs/02-iaas/default-security-group-rules.md
+++ b/docs/02-iaas/default-security-group-rules.md
@@ -1,0 +1,89 @@
+# Guide: How to handle default Security Group Rules
+
+Security Groups in OpenStack are part of the network security mechanisms and resemble sets of firewall rules.
+Every newly created Security Group may already have a few rules applied to it.
+Those rules are managed under the `default Security Group Rules` endpoint of OpenStack and they can be configured by an administrator.
+
+## How to comply to the standard for default Security Group Rules
+
+The standard for default Security Group Rules does not allow any ingress traffic rules, but keeps egress rules as optional.
+
+To test what default Security Group rules are currently in place this command can be used:
+
+```bash
+openstack default security group rule list
+```
+
+If there are any ingress rules in this set of rules, administrators MUST delete them with the following command:
+
+```bash
+openstack default security group rule delete $DEFAULT_SG_RULE_ID
+```
+
+Egress rules can also be deleted this way, if security guidelines specify it.
+
+### Example of a conform set of default Security Group Rules
+
+This is an exapmle how a table with default Security Group Rules may look like:
+
+```bash
+$ openstack default security group rule list
++--------------------------+-------------+-----------+-----------+------------+-----------+-----------------------+----------------------+--------------------------------+-------------------------------+
+| ID                       | IP Protocol | Ethertype | IP Range  | Port Range | Direction | Remote Security Group | Remote Address Group | Used in default Security Group | Used in custom Security Group |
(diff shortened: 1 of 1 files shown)


berendt:
@markus-hentsch Can you please review this one?

docs/02-iaas/default-security-group-rules.md
```diff
@@ -0,0 +1,89 @@
...
+
+```bash
+openstack default security group rule delete $DEFAULT_SG_RULE_ID
+```
+
+Egress rules can also be deleted this way, if security guidelines specify it.
+
+### Example of a conform set of default Security Group Rules
+
+This is an exapmle how a table with default Security Group Rules may look like:
```
markus-hentsch, line 27:
```suggestion
This is an example how a table with default Security Group Rules may look like:
```

docs/02-iaas/default-security-group-rules.md
```diff
@@ -0,0 +1,89 @@
...
+## Other possible configuration options with default Security Group Rules
+
+:::tip
+
+Creating default Security Group Rules works in a similiar way as creating rules for one Security Group.
+The parameters that can be used to describe rules are mostly the same.
+
+:::
+
+The default security Group Rules can differ between default Security Groups and all other Security Groups - called `custom Security Group`.
```
markus-hentsch, line 50:
```suggestion
Default Security Group Rules may be adjusted individually for one of two different types of Security Groups:

1. The default Security Groups that are automatically created for projects initially.
2. Custom Security Groups created by users and initialized with a default rule set.

The default rule set is configured separately for each of those two types.
```

I found the original sentence a bit hard to grasp due to the meaning of "default" and "custom" being separate between rules and groups making this multi-dimensional but hard to convey in a single sentence. That's the reason I rephrased this to make a more clear distinction.

docs/02-iaas/default-security-group-rules.md
```diff
@@ -0,0 +1,89 @@
...
+So it is possible as an administrator to define rules only for one of those groups of Security Groups.
```
markus-hentsch, line 51:
```suggestion
So it is possible as an administrator to define rules only for one of those types of Security Groups.
```

markus-hentsch (changes requested):

mbuechse:
@josephineSei @markus-hentsch This seems to have been in limbo for weeks. What's the status here?

josephineSei:
This belonged to https://github.com/SovereignCloudStack/standards/issues/521. Taht issue was closed automatically, when the standard was merged. So this PR was not on my focus anymore.
But as we have a standard document and within that document, we advise to use the current openstack defaults. I don't think we need a guide anymore. What do you think @markus-hentsch ?

markus-hentsch:
> This belonged to [SovereignCloudStack/standards#521](https://github.com/SovereignCloudStack/standards/issues/521). Taht issue was closed automatically, when the standard was merged. So this PR was not on my focus anymore. But as we have a standard document and within that document, we advise to use the current openstack defaults. I don't think we need a guide anymore. What do you think @markus-hentsch ?

I agree.

We already added a [user-facing guide](https://docs.scs.community/docs/iaas/guides/user-guide/openstack/security-groups) for custom SG rule handling.
This PR here was initially intended to add an operator-facing guide for the SG rule defaults but since the corresponding standard ended up not deviating from upstream in that regard, all of this can easily be deduced from the standard itself in combination with the upstream documentation of `openstack default security group rule ...`.

PR #168, opened 2024-03-20
Feat/add relnotes r6

garloff:


garloff:
Unclean git history. Superceded by #169.

PR #172, opened 2024-03-25
Fix correct repo link

maxwolfs:


Belgeron (approved):
Now it points to the correct repo

PR #173, opened 2024-03-26
Create mandatory-and-supported-os-services.md

josephineSei:
This is the first draft for the OpenStack services lists.
related to https://github.com/SovereignCloudStack/issues/issues/528


docs/02-iaas/mandatory-and-supported-os-services.md
```diff
@@ -0,0 +1,56 @@
...
+The following services MAY be present in SCS-compliant IaaS-Deployment and their integration had been tested by the SCS:
+
+| OpenStack Service | description | comment |
+|-----|-----|-------|
+| **Barbican** | Key Manager service | maybe a future mandatory service |
+| **Ceilometer** | Telemetry service | |
+| **Designate** | DNS service | |
+| **Gnocchi** | Time Series Database service | needed for Ceilometer |
+| **Heat** | Orchestration service | **maybe unsupported?** |
+| **Horizon** | Dashboard | |
```
fkr, line 37: From my understanding the Iaas Reference Implementation is supposed to be able to host the KaaS reference implementation out of the box. and that depends on Octavia being present (from my understanding), so having Octavia around is a must.

docs/02-iaas/mandatory-and-supported-os-services.md
```diff
@@ -0,0 +1,55 @@
...
+| OpenStack Service | description |
+|-----|-----|
+| **Cinder** | Block Storage service |
+| **Glance** | Image service |
+| **Keystone** | Identity service |
+| **Neutron** | Networking service |
+| **Nova** | Compute service |
+| **Octavia** | Load-balancer service |
+| **Placement** | Hardware Describing Service for Nova |
+| **S3 API object storage** | No formal standard exists, many implementations: Swift, RadosGW, minio... |
```
artificial-intelligence, line 24: note: the reader might wonder why this is included when there is no real standard for S3 APIs and the existing implementations all slightly differ.

I think it's safe to say that it is still included because of it's general usefulness and for many usecases the implementations are similar enough to maybe not matter.

It might be worth it though, to add some kind of warning for powerusers, that they should research if their S3 feature is supported by the used implementation.

That being said: Would it make sense to mandate, that the used S3 Implementation must be at least publicly documented so users _can_ actually  research if the implementation is fit for their purpose?

I think that could be important, at least for power users and should not put too much of a burden on the CSP. Maybe this can also already be discovered via existing APIs within S3 itself? But I really don't know about the latter part.

docs/02-iaas/mandatory-and-supported-os-services.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Supported OpenStack services
+
+The following services MAY be present in SCS-compliant IaaS-Deployment and their integration had been tested by the SCS:
```
artificial-intelligence, line 28: what does "their integration had(sic!) been tested" mean here exactly?
e.g. I'm not aware of any automated tests we run for masakari (maybe I just don't know them, it's hard to track every single test across different repositories).

docs/02-iaas/mandatory-and-supported-os-services.md
```diff
@@ -0,0 +1,55 @@
...
+| OpenStack Service | description | comment |
+|-----|-----|-------|
+| **Barbican** | Key Manager service | maybe a future mandatory service |
+| **Cloudkitty** | Rating/Billing service | maybe move to unsupported |
+| **Ceilometer** | Telemetry service | |
+| **Designate** | DNS service | |
+| **Gnocchi** | Time Series Database service | needed for Ceilometer |
+| **Heat** | Orchestration service | **maybe unsupported?** |
+| **Horizon** | Dashboard | |
+| **Ironic** | Bare Metal libraries | |
```
artificial-intelligence, line 39: Ironic is not only a library but also provides actual services, imho this fits it better (stolen from the ironic docs)?
```suggestion
| **Ironic** | Bare Metal provisioning service | |
```

docs/02-iaas/mandatory-and-supported-os-services.md
```diff
@@ -0,0 +1,55 @@
...
+| **Manila** | Shared File Systems service | |
+| **Masakari** | Instances High Availability service | |
+| **Skyline** | Dashboard | if not used, move to unsupported |
+
+## Unsupported OpenStack services
+
+All other OpenStack services that are not mentioned in the mandatory or supported lists are not tested for their integration and behavior by the SCS.
```
artificial-intelligence, line 46: nit:
```suggestion
All other OpenStack services that are not mentioned in the mandatory or supported lists are not tested for their integration and behavior by the SCS community.
```

artificial-intelligence (changes requested):

docs/02-iaas/mandatory-and-supported-os-services.md
```diff
@@ -0,0 +1,55 @@
...
+| **Nova** | Compute service |
+| **Octavia** | Load-balancer service |
+| **Placement** | Hardware Describing Service for Nova |
+| **S3 API object storage** | No formal standard exists, many implementations: Swift, RadosGW, minio... |
+
+## Supported OpenStack services
+
+The following services MAY be present in SCS-compliant IaaS-Deployment and their integration had been tested by the SCS:
+
+| OpenStack Service | description | comment |
```
markus-hentsch, line 30: I think we should get rid of the comment column altogether in this document. The remarks in there sound too informal for an offical SCS document (as in docs page) and more like internal notes in my opinion. We could add these remarks in a more elaborated form within "Design Considerations" of a standard for SovereignCloudStack/issues#528 instead.

docs/02-iaas/mandatory-and-supported-os-services.md
```diff
@@ -0,0 +1,55 @@
...
+| **Masakari** | Instances High Availability service | |
+| **Skyline** | Dashboard | if not used, move to unsupported |
+
+## Unsupported OpenStack services
+
+All other OpenStack services that are not mentioned in the mandatory or supported lists are not tested for their integration and behavior by the SCS.
+Those services MAY be used in IaaS-Deployments but the SCS will not assume they are present and potential issues that occur during the deploying or usage have to be handled by the CSP alone.
+
+There are still some notable services, that might come into the supported list, when there will be enough CSPs wanting this.
+This list is not exhaustive:
```
markus-hentsch, line 50: I suggest rephrasing this a bit for two reasons:

1. emphasize that the SCS standards do not offer any guarantees regarding compatibility/reliability more strongly
2. do not suggest that any of those services may get the supported state if enough CSPs beg for it, it sounds too informal and does not belong in this informational document, in my opinion
    - if you still want to communicate the mutable nature of this, we could consider adding a sentence like "Feel free to reach out to the SCS community to discuss adding new services to the supported list" or something along those lines

Here's my suggested rephrase:

```suggestion
Those services may be integrated into IaaS deployments by a CSP on their own responsibility but the SCS will not assume they are present and potential issues that occur during deployment or usage have to be handled by the CSP on their own accord.
The SCS standard offers no guarantees for compatibility or reliability of services categorized as unsupported in conjunction with an SCS-conformant infrastructure.

Some notable but currently unsupported services which address popular use cases are listed below (list not exhaustive):
```

docs/02-iaas/mandatory-and-supported-os-services.md
```diff
@@ -0,0 +1,55 @@
...
+
+| OpenStack Service | description | comment |
```
markus-hentsch, line 52: As commented above for the other table, I suggest removing the comment column and moving any statements into the Design Considerations section of the corresponding standard document.

markus-hentsch (changes requested):

josephineSei:
closing this in favor of https://github.com/SovereignCloudStack/standards/pull/587

Issue #113, opened 2023-11-09
Upgrade docusaurus to version 3


maxwolfs:
See the Migration Guide: https://docusaurus.io/docs/migration/v3

It will have some breaking changes.

Upgrade complete but the build is still failing due to some remaining errors that must be fixed in other repositories. See the summary:

## MDX Compilation Errors

- [x] [standards](https://github.com/SovereignCloudStack/standards)
- [x] [k8s-cluster-api-provider](https://github.com/SovereignCloudStack/cluster-stack-provider-openstack)
- [x] [release-notes](https://github.com/SovereignCloudStack/release-notes)

### Components

**k8s-cluster-api-provider:**

`doc/Upgrade-Guide.md `(line 111, column 20): Unexpected /, expected a name-starting character.
`doc/requirements.md` (line 15, column 24): Unexpected /, expected a name-starting character.

**Releases**

`ReleaseX.md` (line 4, column 39): Unexpected Y after self-closing slash, expected >.
`Release1.md `(line 35, column 8): Unexpected /, expected a name-starting character.
`Release4.md `(line 78, column 160): Unexpected /, expected a name-starting character.

**Standards**

`scs-0100-v1-flavor-naming.md` (line 92, column 76): Unexpected 2, expected a name-starting character.
`scs-0100-v2-flavor-naming.md` (line 121, column 76): Unexpected 2, expected a name-starting character.
`scs-0100-v3-flavor-naming.md `(line 129, column 76): Unexpected 2, expected a name-starting character.
`scs-0100-w1-flavor-naming-implementation-testing.md` (line 36, column 8): Unexpected /, expected a name-starting character.
`scs-0104-v1-standard-images.md` (line 107, column 8): Unexpected /, expected a name-starting character.
`scs-0200-v1-using-sonobuoy-for-kaas-conformance-tests.md` (line 52, column 60): Unexpected /, expected a name-starting character.
`scs-0300-v1-requirements-for-sso-identity-federation.md` (line 247, column 10): Unexpected /, expected a name-starting character.
`scs-0400-v1-status-page-create-decision.md` (line 159, column 25): Unexpected /, expected a name-starting character.

For more details on these errors, please refer to the [Upgrade Documentation.](https://docusaurus.io/docs/migration/v3)

PR #175, opened 2024-03-27
Upgrade Docusaurus to V3 and fixes

maxwolfs:
Upgrade and relevant fixes to the new md rules coming with it. This is only fixing the markdown errors with the markdown files within this repository. There are still 14 compilation that prevent from building. Those have to be fixed in other repositories. See parent issue #113


community/tools/github/dco-and-licenses.md
```diff
@@ -44,7 +44,7 @@ line to commit messages.
 ```bash
 My fancy commit message
 
-Signed-off-by: Christian Berendt <berendt@betacloud-solutions.de>
+Signed-off-by: Christian Berendt <$berendt@betacloud-solutions.de>
```
mbuechse, line 47: I can live fine with most changes, but this seems wrong.

docs/glossary.md
```diff
@@ -6,7 +6,7 @@ glossary is not intended to evaluate or standardize specific terms.
 
 ## Definition of a Region
 
-An OpenStack/SCS region consists of at least one or more Availability Zones that share a Control Plane with their services. As a result, they share one API. Also a Control Plane can share one CEPH cluster over different fire departments or each Availbility Zone can have its own CEPH cluster. Within the region, any Layer 2 networks are available to the user. Availbility Zones which build a region are connected by redundant low-latency (<2ms (guess!!) ) high bandwidth (10s of Gbps) connections.
+An OpenStack/SCS region consists of at least one or more Availability Zones that share a Control Plane with their services. As a result, they share one API. Also a Control Plane can share one CEPH cluster over different fire departments or each Availbility Zone can have its own CEPH cluster. Within the region, any Layer 2 networks are available to the user. Availbility Zones which build a region are connected by redundant low-latency < 2ms high bandwidth (10s of Gbps) connections.
```
mbuechse, line 9: I think the parentheses should be added back in, to be consistent with the bandwidth. Speaking of consistency though:

```suggestion
An OpenStack/SCS region consists of at least one or more Availability Zones that share a Control Plane with their services. As a result, they share one API. Also a Control Plane can share one CEPH cluster over different fire departments or each Availbility Zone can have its own CEPH cluster. Within the region, any Layer 2 networks are available to the user. Availbility Zones which build a region are connected by redundant low-latency (< 2 ms) high-bandwidth (10s of Gbps) connections.
```

community/tools/github/dco-and-licenses.md
mbuechse: Screenshot from 2024-04-02 17-28-28
No, this is a code segment, so the escape character will not be interpreted. Why did you put it in there?

mbuechse (approved):

package.json
```diff
@@ -37,33 +37,35 @@
     "format": "prettier --log-level warn --write \"./**/*.{jsx,js}\""
   },
   "dependencies": {
-    "@docusaurus/core": "^2.4.3",
-    "@docusaurus/plugin-client-redirects": "^2.4.3",
-    "@docusaurus/preset-classic": "^2.4.3",
-    "@docusaurus/theme-mermaid": "^2.4.3",
-    "@easyops-cn/docusaurus-search-local": "^0.33.6",
+    "@docusaurus/core": "^3.1.1",
```
berendt, line 40: Docusaurus 3.2.0 is out now. I think it makes sense to change this before the merge of this PR.

PR #176, opened 2024-03-27
Add user data backup guide

markus-hentsch:
Part of [#541](https://github.com/SovereignCloudStack/standards/issues/541)


docs/04-operating-scs/08-backups/user_data_backups.md
```diff
@@ -0,0 +1,303 @@
...
+
+The resulting IDs can be used to retrieve the corresponding key using the [Barbican instructions](#barbican-secrets-backup-using-download) above.
+
+:::caution
+
+Note that the key retrieved from the secret is not immediately usable as LUKS passphrase to the image data of the volume.
+OpenStack does some processing to the key before it is passed to the LUKS encryption, which must be mimicked accordingly in order to unlock the encryption outside of OpenStack!
+
+See the [example procedure for converting the LUKS key](#luks-encryption-key-conversion-to-decrypt-volume-images) in the appendix section.
+
```
josephineSei, line 236: Would it work, when you upload both: the image with similar metadata and the key into another OpenStack deployment?


I know that it most likely will fail with the image upload, but this is somehow interesting.

docs/04-operating-scs/08-backups/user_data_backups.md
```diff
@@ -0,0 +1,303 @@
...
+### Image creation action for servers with attached volumes
+
+When the `createImage` action of the Compute API (`openstack server image create`) is used on virtual machines that have at least one volume attached, a snapshot will be created for each attached volume individually and referenced in the resulting image's metadata.
+
+This happens regardless of whether the virtual machine has an Ephemeral Storage disk attached.
+In case of an Ephemeral Storage disk, only the Ephemeral Storage is copied into the Glance image as a 1:1 copy.
+
+In case of a virtual machine that has no Ephemeral Storage but only volumes, the `createImage` action leads to a Glance image that only consists of metadata (including the resulting volume snapshot references) but carries no actual binary data.
+
+![Figure: createImage action flow involving Ephemeral Storage and/or volumes](./images/user_data_backups_figure1.png)
```
josephineSei, line 250: Why is the shelve command integrated there?
For ephemeral storage the procedure is alike, but I don't know whether the metadata of the image is the same for both.
In case of the volume it will not be touched for the shelve action.

josephineSei (commented):
I was a bit confused by the graphic, and did not see any mention of "shelve" in this guide.

docs/04-operating-scs/08-backups/user_data_backups.md
markus-hentsch: Good point. Maybe I need to be a bit more verbose about the shelve operation in the rest of the documentation. Shelve could be considered some form of (at least temporary) backup from the user perspective. Since the same (imho) non-obvious behavior as for `server image create` regarding mixed ephemeral and volume storage applies for shelve, I wanted to be as clear as possible in the user guide.

docs/04-operating-scs/08-backups/user_data_backups.md
markus-hentsch: Very intriguing point. I haven't thought about that yet. I could see this working via `openstack secret store ...` + `openstack image create --property cinder_encryption_key_id=$SECRET_ID ...`.

I will test that.

On that note... should the guide also cover backup restore procedures aside from manually cracking open the LUKS dump locally as currently illustrated in the appendix? I guess it would be a good idea now that you bring it up.

josephineSei: Well I see this use case:
1. Download image from one deployment (maybe encrypted)
1.1. Download associated key
2. Upload Image to another deployment either as backup or already ready to use
2.1. Upload associated key

as more common than storing a downloaded image privately.

Describing a tested and safe way for that would be also good I think.

markus-hentsch:
(converting this back to draft for the time being as I finish up adding proper backup restore instructions)

docs/04-operating-scs/08-backups/user_data_backups.md
markus-hentsch: I have now included restore instructions addressing all of this.

docs/04-operating-scs/08-backups/user_data_backups.md
markus-hentsch: Actually, you are right. I checked again and the behavior differs from `image create`. No snapshots are taken of volumes and no image is created if there is no Ephemeral Storage. I don't think shelve could count as any form of backup after all.

I have removed the references to shelve from the figure. Thanks for bringing this up!

markus-hentsch:
I've added restore instructions for all mentioned backups now. Putting this back into ready for review again.

docs/04-operating-scs/08-backups/user_data_backups.md
```diff
@@ -0,0 +1,479 @@
+---
+title: User Data Backups
+type:
```
anjastrunk, line 3: Could we write "Guide" here or are only standardizes labels allowed?

```suggestion
type: Guide
```

docs/04-operating-scs/08-backups/user_data_backups.md
```diff
@@ -0,0 +1,479 @@
...
+When creating images from volumes with a volume type that uses encryption, the resulting image will contain the raw LUKS-encrypted blocks of the volume.
+When transferred outside of the IaaS infrastructure, this data is only useful as a backup together with the corresponding encryption key.
+
+Such images can be identified by an attribute called `cinder_encryption_key_id` in the `properties` metadata field of the image.
+It only exists for encrypted images and references the encryption key in Barbican.
+Refer to the [Barbican secrets section](#barbican-secrets-backup-using-download) for instructions on how to backup the key.
+
+:::
+
+The API or the client may be used to initiate the download, for example:
```
anjastrunk, line 58:
```suggestion
The API or the openstack client may be used to initiate the download, for example:
```

docs/04-operating-scs/08-backups/user_data_backups.md
```diff
@@ -0,0 +1,479 @@
...
+
+:::caution
+
+When using the `createImage` Compute API action (e.g. via the `openstack server image create` command) on a virtual machine that has volumes attached to it in addition to its Ephemeral Storage disk, the volumes will not be backed up into the image. Instead, a snapshot will be created for each attached volume and referenced in the image metadata. This does not replace genuine volume backups.
+
+See the [corresponding appendix section](#image-creation-action-for-servers-with-attached-volumes) for further details.
+
+:::
+
+Ephemeral Storage disks of virtual machines can be backed up to Glance images easily by using the `createImage` Compute API action or the corresponding client command:
```
anjastrunk, line 76:
```suggestion
Ephemeral Storage disks of virtual machines can be backed up to Glance images easily by using the `createImage` Compute API action or the corresponding openstack client command:
```

docs/04-operating-scs/08-backups/user_data_backups.md
```diff
@@ -0,0 +1,479 @@
...
+
+It might be difficult or even impossible for a user to transfer backups created by the Cinder Backup API outside of the IaaS infrastructure, depending on the backup backend.
+A more easily accessible backup of volumes can be created by using Glance images.
+See the [section about volume data backup using Glance images](#volume-data-backup-using-glance-images) for details.
+
+:::
+
+### Backup of detached volumes
+
+Backups can be created using the Cinder Backup API or the corresponding client commands:
```
anjastrunk, line 110:
```suggestion
Backups can be created using the Cinder Backup API or the corresponding openstack client commands:
```

docs/04-operating-scs/08-backups/user_data_backups.md
```diff
@@ -0,0 +1,479 @@
...
+
+### Backup of attached volumes
+
+:::note
+
+When creating backups of attached (in-use) volumes, the `force` parameter has to be used. These backups capture a state of the full volume at runtime. They will be crash-consistent but not app-consistent.
+
+:::
+
+In case of attached (in-use) volumes, backups can only be created while also specfiying the `force` parameter:
```
anjastrunk, line 130: This sentence is a repetition of the sentence above. I would re-write it. e.g. as follows:
```suggestion
The API or the client may be used to backup attached volume,  for example:
```

anjastrunk (approved):
Very good guide to make and restore user backup. Specially, I like the way each chapter is written. It allows he reader to jump in a specific section and getting all relevant information, without reading the guide as a whole.

docs/04-operating-scs/08-backups/user_data_backups.md
markus-hentsch: I rephrased the note banner instead.

docs/04-operating-scs/08-backups/user_data_backups.md
markus-hentsch: I looked at other docs pages and this header does not seem mandatory/necessary at all. I replaced it with a single Markdown headline instead.

markus-hentsch:
I moved the page from "Operating SCS > Guides" to "IaaS Layer > Guides > User Guide" where the new security group best practice guide also resides. I think this is a better location since this guide is addressed to users first and foremost, not operators.

| Old Location | New Location |
|---|---|
| old_pos | new_pos |

anjastrunk (approved):

PR #178, opened 2024-04-04
Add note to Domain Manager guide about Identity Federation

markus-hentsch:
This came up in a past IAM community call. Domain Manager functionality is unavailable for Identity Federation due to identities being managed outside of Keystone in this case.

This PR adds a small note to beginning of the Domain Manager guide to clarify this limitation.


gtema (commented):
honestly speaking I am not sure about that. It is very depending on how the federation is done. There is nothing preventing making a federation, landing a federated user in Keystone and admin granting role to it. Moreover, in the case of using domain-driver for the Keycloak integration there is no possibility to have local users at all with all users/groups coming directly from Keycloak. As said - it depends.

markus-hentsch:
> honestly speaking I am not sure about that. It is very depending on how the federation is done. There is nothing preventing making a federation, landing a federated user in Keystone and admin granting role to it. Moreover, in the case of using domain-driver for the Keycloak integration there is no possibility to have local users at all with all users/groups coming directly from Keycloak. As said - it depends.

I see, thanks for clarifying. I will close this PR then. If the situation is not as clear cut, we should avoid jumping to wrong conclusions in the documentation.

PR #182, opened 2024-04-23
Extend kaas playgrounds

jschoone:
Hi @frosty-geek,
the SCS Community Hackathon was a boost on developing and using various projects with the help of the kaas-playground projects on gx-scs. We still actively use those and would appreciate if we can extend them until the end of the project funding.


frosty-geek (approved):
LGTM

Issue #185, opened 2024-04-29
Extend general guidelines of conformance tests wrt to regression/unit tests


martinmo:
Extend the guidelines created in #154 with information on regression/unit tests for the conformance tests (see https://github.com/SovereignCloudStack/standards/issues/447#issuecomment-1968905747).

For completeness, the guidelines should also mention that the conformance checks should be executable with the least privileges possible, i.e., non-admin credentials for OpenStack and Kubernetes APIs and of course the scripts should never need root priveleges (this is already somewhat documented/standardized with `scs-0004-v1`).

PR #186, opened 2024-04-29
Extend test guidelines with implementation notes for unit/regression tests

martinmo:
Work-in-progress.

Resolves: #185

### DIFF
--- a/contributor-docs/operations/tests/implementation-notes.md
+++ b/contributor-docs/operations/tests/implementation-notes.md
@@ -1,0 +1,74 @@
+---
+title: Implementation Notes for SCS Conformance Tests
+type:
+status: Draft
+track: Global
+---
+
+SovereignCloudStack (SCS) uses [conformance tests][gh-scripts] to certify
+compliance of a given cloud offering with respect to a given [certificate
+scope][cert-scopes] such as *SCS Compatible IaaS v4*.
+Our aim is that these tests are reliable, consistent and comprehensible for
+the users.
+This document summarizes the current best practices to achieve this goal.
+
+
+## 1. RFC2119 Keywords
+
+Creating a conformance test for a standard involves the correct mapping of
+the keywords such as MUST and SHOULD, which is described in the
+[SCS RFC2119 Keyword Test Guide][scs-rfc2119-guide].
+
+
+## 2. Unit and Regression Tests
+
+Conformance tests will inevitably contain some non-trivial algorithms, be it for
+parsing flavor names or evaluating CVE vulnerability reports.
+Developers of such tests are therefore *strongly* encouraged to include unit
+and regression tests for the conformance test's logic.
+As a bonus, a well written unit tests also makes it easier for reviewers to
+cross check a given pull request with new or enhanced conformance tests.
+
+### Requirements
+
+* Developing and running unit tests requires setting up the environment as outlined
+  in the [README of the SCS conformance tests][tests-readme].
+* The name of the Python module implementing the conformance check should use be
+  a valid [Python identifier][python-identifiers], to make it possible to import
+  module members with the `import` statement.
+* Any module that ends in `_test`, will be picked up by pytest in our CI pipeline
+  (in other words, all files matched by the pattern `Tests/**/*_test.py`).
+  Pytest will execute any function prefixed with `test_` as a test function, or,
+  alternatively, tests based on the `unittest` module from the Python stdlib.
+
+### Testable Conformance Tests
+
+* Divide conformance tests scripts into smaller, loosely coupled units, i.e.,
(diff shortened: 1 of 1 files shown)


contributor-docs/operations/tests/test-implementation-guide.md
```diff
@@ -0,0 +1,147 @@
...
+RFC2119 Keyword Test Guide][scs-rfc2119-guide].
+
+## 2. Unit and Regression Tests
+
+Test authors are *strongly* encouraged to include unit and regression tests for
+the conformance test's logic.
+Conformance tests will inevitably contain some non-trivial algorithms, be it for
+parsing flavor names or evaluating CVE vulnerability reports.
+As a bonus, a well written unit test also makes it easier for reviewers to
+cross check a given pull request with new or enhanced conformance tests.
```
mbuechse, line 29: This is a non sequitur. I think a linking sentence is missing. The sentence before only states a problem without saying why having a unit test is good (which is kind of self evident whn you think about it, but for the flow of the text, it should be added).

contributor-docs/operations/tests/test-implementation-guide.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Write Testable Conformance Tests
+
+Software design fills complete book shelves and this section can only cover some
+basic best practices.
+As a general rule of thumb, the earlier you begin writing unit tests, the better
+– it will force you to write modules that are easy to test.
+
+Divide conformance tests scripts into smaller, loosely coupled units, i.e.,
+functions that serve one purpose.
```
mbuechse, line 63:
```suggestion
functions that serve one purpose each.
```

contributor-docs/operations/tests/test-implementation-guide.md
```diff
@@ -0,0 +1,147 @@
...
+Pass externally retrieved data as function arguments instead of hardcoding the
+calls to the retrieval functions inside a function.
```
mbuechse, line 65: Prefix this by saying: `Practice inversion of control, that is to say, ` (or something similar)

contributor-docs/operations/tests/test-implementation-guide.md
```diff
@@ -0,0 +1,147 @@
...
+    # using sys.exit(…) only here makes it possible to unit test main(…)
+    sys.exit(main(sys.argv))
+```
+
+Adhering to this style makes it easier to test an algorithm in isolation,
+without actually making a call to some external service.
+
+### Pytest Test Example
+
+Assuming we want to unit test some members of the module `property_compliance`, we would
```
mbuechse, line 106:
```suggestion
Assuming we want to unit test some members of the module `property_compliance` from the previous section, we would
```

mbuechse (commented):
Apart from a few little nit picks, I really like this. I'm only wondering why we now have two documents, and how developers will know what document to read for which purpose. I have the feeling that they need to read both documents anyway in order to do meaningful work. Can they be merged in a reasonable way?

martinmo:
> I'm only wondering why we now have two documents, and how developers will know what document to read for which purpose. I have the feeling that they need to read both documents anyway in order to do meaningful work. Can they be merged in a reasonable way?

The other document specializes on the RFC keywords and without a complete restructuring I didn't see how I could fit my topics in there. For this reason, I decided to create a separate document. The new document can be seen as the more general entrypoint which a newcomer should read first.

I agree with you that this isn't obvious at the moment. One more thing which isn't obvious: the testing related contributor documents are currently placed below the "operations" hierarchy:

```
contributor-docs/
├── index.md
└── operations/
    ├── iam/
    │   ├── identity-federation-in-scs.md
    │   └── openstack-federation-via-oidc.md
    ├── operations/
    │   └── zuul-ci-cd-quickstart-user-guide.md
    └── tests/
        ├── rfc2119-keyword-test-guide.md
        └── test-implementation-guide.md
```

My suggestion: create a new "development" hierarchy and move "tests" there. Furthermore, solve the "which document to read first?" question by putting an `index.md` in there. So it will look like this:

```
contributor-docs/
├── development/
│   ├── index.md
│   └── tests/
│       ├── rfc2119-keyword-test-guide.md
│       └── test-implementation-guide.md
├── index.md
└── operations/
    ├── iam/
    │   ├── identity-federation-in-scs.md
    │   └── openstack-federation-via-oidc.md
    └── operations/
        └── zuul-ci-cd-quickstart-user-guide.md
```

@mbuechse what do you think about this proposal?

martinmo:
We (@cah-patrickthiem, @mbuechse and me) just had a discussion about the above proposal and it was agreed to proceed with it.

mbuechse (approved):

PR #190, opened 2024-05-07
Added initial documentation about example setup

JuanPTM:
initial version of the example setup description

### DIFF
--- a/docs/05-iam/SCS-example-setup-configuration-description.md
+++ b/docs/05-iam/SCS-example-setup-configuration-description.md
@@ -1,0 +1,21 @@
+# Example setup configuration in SCS deployment explained
+
+The following document explains the idea behind the example configuration is done.
+The playbook creates a proxy realm used to connect with `Keystone`, a customer realm, the clients needed to connect the realms, the identity brokering for the customer realm, a login flow
+to be able to use the `home-IdP-discovery` plugin to redirect to the correct customer realm and an example user.
+SCS operators can find the playbook that creates the setup [here](https://github.com/osism/ansible-playbooks/blob/main/playbooks/infrastructure/keycloak-oidc-client-config.yml).
+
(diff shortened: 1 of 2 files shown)


reqa:
Please add these points:
* What does this describe? A link to the playbook/task would probably be good?
* Motivation: For which purpose is this interesting? Mention the SCS persona that is the target of this documentation (see https://github.com/SovereignCloudStack/standards/blob/773646d1d6af19b151b21e6b706c946e8169c07d/GLOSSARY.md#personas)

garloff:
Much better now.
Where on the docs page should this go?

JuanPTM:
> Much better now. Where on the docs page should this go?

I don't know, maybe @maxwolfs can help us here

maxwolfs:
> Example federation setup (testbed)

I would put it in IAM/Guides/Example federation setup (testbed) next to Domain Manager setup and usage.

Then it has to be added in the sidebar here: https://github.com/SovereignCloudStack/docs/blob/main/sidebarsDocs.js#L357

JuanPTM:
> > Example federation setup (testbed)
> 
> I would put it in IAM/Guides/Example federation setup (testbed) next to Domain Manager setup and usage.
> 
> Then it has to be added in the sidebar here: https://github.com/SovereignCloudStack/docs/blob/main/sidebarsDocs.js#L357

Sidebar file changed and original file moved to the correct location, 

how is it looking now

maxwolfs:
Could you please work on a local branch within the main repository instead of using a fork? Then the build will work and this will also allow us to review the changes on the rendered staging branch directly on https://docs-staging.scs.community/

JuanPTM:
I've created another MR using a local branch instead of a fork. 

https://github.com/SovereignCloudStack/docs/pull/202

PR #203, opened 2024-06-14
Sync compliance table with README.md in standards repo

mbuechse:
Just copy&paste from current revision in standards repo


gtema (approved):

mbuechse:
Note: the failing link check can in this case be ignored, because it's just some temporary downtime of the gx-scs health monitor.

PR #209, opened 2024-06-28
Transfer current cloud table from standards repo

mbuechse:
- bump gx-scs v2 -> v3
- aov.de -> www.aov.de
- move syseleven v4 -> v3


gtema:
independent on the check failures: I am fine, but as mentioned I find confusing to see few identical badges in a column. Knowing about virtual rows for regions I understand how to interpret it, but otherwise it is hard. And since MD doesn't support merged cells I suggest either adding a text in front of a badge with a region name or include it into the badge title directly

mbuechse:
Yes, @gtema this shall be improved, but it's a separate issue.

mbuechse:
Oh, and the checks... the Health Monitor for gx-scs still seems to be unreachable. Apart from that, the failures are not due to changes of this PR.

gtema (approved):

PR #210, opened 2024-06-28
Extend cert scope tables to include parameters

### LINKED ISSUE #208: Extend cert scope tables to include parameters
Now that formal parameters are a thing, we need to document them (documentation was the main reason to introduce the concept in the first place).

mbuechse:
resolves #208


maxwolfs (approved):
LGTM!

Btw, const `state` is currently not being used

PR #211, opened 2024-07-11
Add color coding to the to standards table in the docs page

cah-patrickthiem:
Added the color mapping with a react component. Needed to change the overview.md file to .mdx format.

related issue: [445](https://github.com/SovereignCloudStack/standards/issues/445)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 /docs/06-releases
 /standards/*.md
 /standards/*/*.md
+/standards/*/*.mdx
 /standards/scs-*.yaml
 
 # Dependencies
--- a/populateStds.js
+++ b/populateStds.js
@@ -7,15 +7,13 @@ Standards are the core deliverable of SCS. By standardizing the open source soft
 
 SCS standards are discussed, developed and maintained in the community by the corresponding teams (see Track in the table below), which naturally include existing users of SCS.`
 const trackIntros = {
-  Global:
-    'This track encompasses the foundational standards that guide the overall structure, documentation, and general topics related to the Sovereign Cloud Stack. It serves as the core framework, ensuring consistency, clarity, and comprehensibility across all aspects of the cloud stack, fostering an environment where information is easily accessible and understood.',
-  IaaS: 'The IaaS Layer Standards track focuses on the protocols, guidelines, and specifications that govern the infrastructure as a service layer. This encompasses standards for virtual machines, storage, networking, and other foundational resources, ensuring seamless, efficient, and secure operation, interoperability, and management of the underlying cloud infrastructure.',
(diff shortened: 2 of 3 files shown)


cah-patrickthiem:
Regarding colors and high contrast ratio:
I did use a [contrast checker](https://webaim.org/resources/contrastchecker/) to pick the colors in contrast especially to the link color. I always made sure "WCAG AA" was met in normal text.
To be sure the colors match each other I used "HSL" ratio in the color picker.

The colors I picked are all relatively light pastel colors:
red: deprecated
yellow: draft
green: stable
blue: effective

However, if you wish other colors, please give feedback. Also the stable and effective colors could be swapped imo.

cah-patrickthiem:
looks like this:
image

mbuechse (commented):
Good work! I would like to build something based on this, and I think 95 % of your work can be used for that. I have the following desiderata:

- I have the impression that the colors are being set using plain JavaScript. So I would try to not use React, but onload or something.
- I absolutely hate that the whole `populateStds.js` has been reformatted. I would like to keep reformatting and semantic change separate (using separate PRs). But I don't even see the need for the reformatting.

cah-patrickthiem:
> Good work! I would like to build something based on this, and I think 95 % of your work can be used for that. I have the following desiderata:
> 
> * I have the impression that the colors are being set using plain JavaScript. So I would try to not use React, but onload or something.
> * I absolutely hate that the whole `populateStds.js` has been reformatted. I would like to keep reformatting and semantic change separate (using separate PRs). But I don't even see the need for the reformatting.

Regarding react: I used react here, because the docusaurus documentation says that if you want to enrich markdown with more features (e.g. coloring), use react: see [here](https://tutorial.docusaurus.io/docs/tutorial-basics/markdown-features/#links:~:text=MDX%20can%20make%20your%20documentation%20more%20interactive%20and%20allows%20using%20any%20React%20components%20inside%20Markdown%3A)
In general it is plain js code, but in order to be parsed by docusaurus, we need that little react "useEffect" function there. Besides that, no further react stuff is used. I tested it also with <script> tags, but that did not work. 
If you have another idea I am not aware of, let me know.

Second, the reformatting was somehow done automatically, I'll try to figure out why and reverse it.

mbuechse:
I probably won't become a fan of Docusaurus, but so be it. Thanks for the explanation!

mbuechse (approved):
LGTM. Just one minor thing: dark mode. Maybe we can postpone dark mode. If you can address it right away, all the better.

cah-patrickthiem:
> LGTM. Just one minor thing: dark mode. Maybe we can postpone dark mode. If you can address it right away, all the better.

Well, I think the colors also work in dark mode, at least regarding the signaling purpose, for sure the colors do not match the dark background in an esthetic way.

I would need to investigate on how to implement this.

Here is a screenshot in dark mode. If that is ok for you right now, we can merge it for now imo. Let me know what you prefer.
image

mbuechse:
I have already approved it, noting that we can postpone dark mode. Go ahead!

PR #218, opened 2024-08-05
Refresh clouds table once again

mbuechse:


gtema (approved):
could you please short description at least into the PR description explaining rationale of this change? I assume it is about consistency, but better to be explicit

PR #220, opened 2024-08-09
Simplify standards overview table

### LINKED ISSUE #215: Simplify standards overview table
See here: https://docs.scs.community/standards/standards/overview

Column “Stable” represents an edge case that almost never happens. Therefore, it’s empty almost all the time.

Merge it with “Effective”, marking future-stable versions with some symbol (such as dagger or asterisk).
Additionally, mark future-deprecated versions with another symbol.

cah-patrickthiem:
Removed the "Stable" column from tables and introduced * markers for future effective documents and † for future deprecated documents.

I also changed the legend according to the new implementations.

closes #215

### DIFF
--- a/populateStds.js
+++ b/populateStds.js
@@ -13,22 +13,12 @@ const trackIntros = {
     'IAM': 'This track revolves around Identity and Access Management (IAM) standards, providing guidelines for ensuring secure and efficient user authentication, authorization, and administration. It addresses issues related to user identity, permissions, roles, and policies, aiming to safeguard and streamline access to cloud resources and services.',
     'Ops': 'Operational Tooling Standards cover the protocols and guidelines associated with tools and utilities used for monitoring, management, and maintenance of the cloud environment. This includes standards for status pages, alerts, logs, and other operational tools, aiming to optimize the reliability, performance, and security of cloud services and resources.',
 }
-const headerLegend =
-  '*Legend to the column headings and entries:\nDocument states: Draft, Effective, Deprecated (and no longer effective).\nEntries in the effective column marked with an * are stable right now but turn to effective documents in the near future.\nEntries in the effective column marked with a † will turn deprecated in the near future.'
+const legend =`   
+<p>*Legend to the column headings and entries:</p>\n
+- Document states: Draft, Effective, Deprecated (and no longer effective)
+- Entries in the effective column marked with an * are stable right now but turn to effective documents in the near future
+- Entries in the effective column marked with a † will turn deprecated in the near future`;
 
-// convert the string into an array of lines
-const legendLines = headerLegend.split('\n')
-
-// extract the first line separately
-const firstLine = legendLines.shift()
-
-// render as a bullet-point list
-const legendWithBullets = `
-<p>${firstLine}</p>
-<ul>
-  ${legendLines.map((line) => `<li>${line}</li>`).join('')}
-</ul>
-`
 const reactHighlightTableCellBackground = `
 import { useEffect } from 'react';
 export const TableCellStyleApplier = () => {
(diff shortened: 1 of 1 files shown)


populateStds.js
```diff
@@ -93,18 +108,6 @@ filenames.forEach((filename) => {
...
-    var lines = []
-    if (fs.existsSync(fn)) {
-        lines = fs.readFileSync(fn, 'utf8').split('\n')
-        var tableIdx = lines.findIndex((line) => line.trim().startsWith('|'))
-        if (tableIdx >= 0) {
-            lines.splice(tableIdx)
-        }
-    } else console.log(`WARNING: file ${fn} not found`)
-    return lines
-}
```
cah-patrickthiem, line 106: When developing on my local computer I noticed that this function is not beneficial, so I excluded it.
The problem was, that the header was not updated properly (the included react code was not executed anymore), therefore I always needed to delete the overview.mdx manually, when I made changes.
I also was not really able to figure out what the purpose is.
If anyone thinks that this function is needed and should stay, please explain why.

populateStds.js
```diff
@@ -13,17 +13,30 @@ const trackIntros = {
...
+// extract the first line separately
+const firstLine = legendLines.shift()
+
+// render as a bullet-point list
+const legendWithBullets = `
+<p>${firstLine}</p>
+<ul>
+  ${legendLines.map((line) => `<li>${line}</li>`).join('')}
+</ul>
+`
```
mbuechse, line 31: I have two questions:

- why do you have legendWithBullets computed when you can just write it down as a literal?
- why don't you use Markdown syntax here?

populateStds.js
mbuechse: I think it's no longer necessary. I can't tell what the problem was in your case. You could supply the non-working `overview.mdx` for further inspection.

populateStds.js
```diff
@@ -176,9 +179,30 @@ ${headerLegend}
...
+                      version:
+                        v.version +
+                        (v.state.futureEffective
+                          ? '*'
+                          : v.state.futureDeprecated
+                          ? '&#8224;'
+                          : '')
+                    }
+                  })
+              ) || '-'
```
mbuechse, line 203: I would prefer to have this logic in mkLinkList itself. This would have multiple advantages:

- this computation here can remain a three-liner
- the version field would not have to be changed (it has a very specific meaning)
- and the asterisk/dagger could be rendered _outside_ the link (as I think it should be)

mbuechse: Oh, and one more advantage: the asterisks/daggers would then also be added for the supplements without changing line 221 (now; formerly 197).

mbuechse (commented):
Overall, good work. I think it does the trick quite well. Still, I ask you to consider my comments.

populateStds.js
cah-patrickthiem: You are right here. I had problems with displaying the first "*" along with the bullet points in plain markdown. But I changed it now to a more plain representation, but still needed to use html to display everything accordingly.

populateStds.js
cah-patrickthiem: Since it is ok from your side to have it excluded, we do not need to track down the problem imo. I already checked it, but was not able to identify any obvious differences. I think there would be the need to check for certain characters and how they are used.

mbuechse: Agreed

populateStds.js
mbuechse: Good point. In any case,I would use `- ` in Markdown for the list.

populateStds.js
cah-patrickthiem: I agree, changed it.

populateStds.js
cah-patrickthiem: For the record, it looks like this:
image

cah-patrickthiem:
For the record, it looks like this:
image

populateStds.js
cah-patrickthiem: done

mbuechse (approved):
Gorgeous!

PR #231, opened 2024-08-21
Fix infra-scan-pipeline missing image

bitkeks:


bitkeks:
@maxwolfs something to have a look at..

jschoone:
We stuck at the same problem which seems to be fixed with this PR

bitkeks:
This is more of a quickfix, we should look at it again with @maxwolfs:

1. The best option is to not limit `source` to `*.md`, but copy the whole directory. But this fails, if one of the Markdown files has issues in rendering
2. Another option is to then give a list of specific files to copy, `"source": ["docs/*.md", "docs/images/*.png"]`. But this will break paths, because the images will be in the same directory level as the .md, breaking relative paths.
3. Third option is to use option 1 but enforcing that included documents are all compatible with Docusaurus. This will always fall back on the people from the upstream repos

PR #242, opened 2024-09-04
feat(security): Add KaaS pentesting tools to docs

bitkeks:
Integrates the KaaS security tooling docs into the docs.

Rename previously added "automated-pentesting" entry to "-iaas" and add "-kaas". Both sections are now under-categories and are both displayed in the component overview.


90n20 (approved):
Lgtm.

PR #252, opened 2024-09-24
Fix/update ws, path-to-regexp, micromatch

garloff:
All found by github...


gtema (approved):

garloff:
This addresses the issues found by dependabot in github.
Note:
- path-to-regexp has a version 8.0.0 (and 0.1.10) that is recommended -- it is however not fully compatible to what the various pieces in docusaurus expect. So I want from 1.x to 1.9.0 and from 2.2.1 to 3.3.0, which results in a succeeding build.
- ws went to 7.5.10 and 8.18.0 as recommended
- minor update to micromatch 

I did regenerate the base64 hashes from the integrity files by downloading the tarballs and calculating the sha512sums myself. There is probably a different official way to do this ...

PR #254, opened 2024-09-24
Update caniuse-lite (browsersl.ist)

garloff:


garloff:
Following the npm deprecations ...

garloff:
I should mention that someone still pulls in old ws.
And these patches here are not complete enough to get rid of stable@0.1.8 nor eslint-config-standard-with-typescript.

gtema (approved):

PR #255, opened 2024-09-25
Update package.lock file

gtema:
Updating dependencies in the package.json silently broke the build_and_deploy workflow where `npm ci` complains on mismatch of versions between packages.json and package-lock.json

Synchronize them using `npm install` which updates the lock file.

Ensure GH `build` action is using same way of building repository as build_and_XXX to ensure this will not happen in future.

Signed-off-by: Artem Goncharov <artem.goncharov@gmail.com>


maxwolfs:
What happened that this is necessary? Seems somehow strange. 🤔

maxwolfs (approved):

PR #260, opened 2024-10-08
Adjust checks: linter and link checker  

maxwolfs:
* remove `standards` repo from being linted in this repo, which leads to unnecessary failing checks
* change onBrokenLinks (docusaurus internal checker during each build process, which is separate to our additional markdown-lint-checker) behavior from `throw` to `warn` to allow relative links from docs of external repo (so even more important to read the logs carefully once a build does not succeed)

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -9,7 +9,7 @@ const config = {
   tagline: 'Documentation and Community Platform for the Sovereign Cloud Stack',
   url: 'https://docs.scs.community',
   baseUrl: '/',
-  onBrokenLinks: 'throw',
+  onBrokenLinks: 'warn',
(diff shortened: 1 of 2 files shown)


scoopex (approved):
LGTM

docusaurus.config.js
```diff
@@ -9,7 +9,7 @@ const config = {
   tagline: 'Documentation and Community Platform for the Sovereign Cloud Stack',
   url: 'https://docs.scs.community',
   baseUrl: '/',
-  onBrokenLinks: 'throw',
+  onBrokenLinks: 'warn',
```
scoopex, line 12: Probably its a good idea to do rigid checking in future.
If the build does not fail, nobody will review the output of the build.

PR #264, opened 2024-10-21
Update additional resource on landing page

maxwolfs:
* fix deployment examples link
* add application examples


michaelbayr (approved):
Das ging flott :)
LGTM

PR #268, opened 2024-10-28
Adding a preinstallation checklist

scoopex:
This list describes some aspects (without claiming to be exhaustive) that should be clarified before a pilot and at least before production installation.

The aim of this list is to reduce:

- Projects that could be more successful
- Long project waiting/implementation times
- Unexpected errors or difficulties
- Major restructuring work soon after the system was initially put into operation
- Unexpected issues that have a major impact on costs.


garloff (approved):
This is a good start, IMVHO.

garloff:
Sidenote on the link-checker CI failure:
- The document references its own source in a request to contribute, which makes sense.
- The doc source does not exist prior to merging this PR
So the failure is harmless.

PR #283, opened 2024-11-27
Restructure Container Docs

jschoone:


fkr:
@jschoone should we still see about getting this into the tree?

PR #285, opened 2024-12-05
Add new logo and icon

maxwolfs:


berendt:
I would prefer to introduce the logo change in advance in the new OSBA forum.

scoopex (approved):
LGTM

PR #292, opened 2025-01-23
Fix/fix faq ciab descr link

garloff:
In the intro "about", the link to CiaB was broken.
Furthermore, it talked about CiaB as a virtual deployment.
While CiaB can be deployed virtually, it is typically not. The point of CiaB is that it's a single-node deployment.

Being at it:
- Fixing the testbed links.
- Move the mentioning that the ref. impl. for virtualization (IaaS) layer is from OSISM ahead of the deployment examples.
- Mention artcodix and HW labs there as well.
- Mention Cluster-Stacks from syself as ref.impl. as well ahead of the description (to have the same structure)

### DIFF
--- a/docs/index.mdx
+++ b/docs/index.mdx
@@ -40,7 +40,7 @@ virtualization.
 
 Read the [testbed docs](/docs/iaas/deployment-examples/testbed) to learn how to get the
 testbed running. Please read carefully through the
-[deployment](https://docs.osism.de/testbed/deployment.html) section of the
+[deployment](https://osism.tech/docs/testbed#deployment) section of the
 manual.
 
 #### Examples for real deployments
--- a/docs/index.mdx
+++ b/docs/index.mdx
@@ -16,26 +16,45 @@ SCS is built, backed, and operated by an active open-source community worldwide.
 
 ## Use Cases and Deployment Examples
 
-### IaaS Layer
+### Virtualization (IaaS) Layer
+
+The SCS IaaS reference implementation is based on [OSISM](https://osism.tech/).
 
 #### Quick Start with Cloud-In-A-Box
 
-The fastest way to get in touch with SCS is to deploy a SCS cloud virtually. The Cloud-In-A-Box was built explicitly for this scenario. Check it out [here](/docs/iaas/guides/deploy-guide/examples/cloud-in-a-box)
+You can do a single node installation for learning, testing or development purposes.
+The Cloud-In-A-Box configuration was built explicitly for this scenario.
+Check it out [here](/docs/iaas/deployment-examples/cloud-in-a-box)
+It comes with a complete set of services, even ceph is part of it
+(despite of course not offering much redundancy on a single-node).
 
 #### Reference Implementation Testbed
+
+The fastest way to get in touch with SCS is to deploy a SCS cloud virtually.
 
 This means that you set up an SCS test installation including all the infrastructure
 pieces such as database, message queueing, ceph, monitoring and logging, IAM, the
 [OpenStack](https://openstack.org/) core services, and (soon) the Container layer
-on top of an existing IaaS platform.
+on top of an existing OpenStack IaaS platform, ideally one that allows for nested
+virtualization.
 
-The SCS IaaS reference implementation is based on [OSISM](https://osism.tech/). Read on the
-[OSISM testbed docs](https://docs.osism.de/testbed/) to learn how to get the
+Read the [testbed docs](/docs/iaas/deployment-examples/testbed) to learn how to get the
 testbed running. Please read carefully through the
 [deployment](https://docs.osism.de/testbed/deployment.html) section of the
 manual.
 
+#### Examples for real deployments
+
+[artcodix](https://artcodix.com/) has [shared](/docs/iaas/deployment-examples/artcodix/)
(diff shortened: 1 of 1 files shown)


garloff:
Cherry-picked to staging in case you want to view the result ...

docs/index.mdx
```diff
@@ -16,27 +16,46 @@ SCS is built, backed, and operated by an active open-source community worldwide.
...
-### IaaS Layer
+### Virtualization (IaaS) Layer
+
+The SCS IaaS reference implementation is based on [OSISM](https://osism.tech/).
 
 #### Quick Start with Cloud-In-A-Box
 
-The fastest way to get in touch with SCS is to deploy a SCS cloud virtually. The Cloud-In-A-Box was built explicitly for this scenario. Check it out [here](/docs/iaas/guides/deploy-guide/examples/cloud-in-a-box)
+You can do a single node installation for learning, testing or development purposes.
+The Cloud-In-A-Box configuration was built explicitly for this scenario.
```
berendt, line 26: Cloud in a Box

docs/index.mdx
```diff
@@ -16,27 +16,46 @@ SCS is built, backed, and operated by an active open-source community worldwide.
...
+Check it out [here](/docs/iaas/deployment-examples/cloud-in-a-box)
+It comes with a complete set of services, even ceph is part of it
```
berendt, line 28: Ceph

docs/index.mdx
```diff
@@ -16,27 +16,46 @@ SCS is built, backed, and operated by an active open-source community worldwide.
...
-on top of an existing IaaS platform.
+on top of an existing OpenStack IaaS platform, ideally one that allows for nested
+virtualization.
 
-The SCS IaaS reference implementation is based on [OSISM](https://osism.tech/). Read on the
-[OSISM testbed docs](https://docs.osism.de/testbed/) to learn how to get the
+Read the [testbed docs](/docs/iaas/deployment-examples/testbed) to learn how to get the
 testbed running. Please read carefully through the
-[deployment](https://docs.osism.de/testbed/deployment.html) section of the
+[deployment](https://osism.tech/docs/testbed#deployment) section of the
```
berendt, line 43: I think it makes more sense to use ` /docs/iaas/deployment-examples/testbed#deployment` here. This way we do not link to an external document.

docs/index.mdx
```diff
@@ -16,27 +16,46 @@ SCS is built, backed, and operated by an active open-source community worldwide.
...
+#### Examples for real deployments
+
+[artcodix](https://artcodix.com/) has [shared](/docs/iaas/deployment-examples/artcodix/)
+some details on their production setup. The SCS team itself has created [extensive
+documentation](/docs/turnkey-solution/hardware-landscape/) including details on the
+used hardware.
+
 ### Container Layer
 
+The Reference Implementation (v2) for the container (Kubernetes-as-a-Services = KaaS) layer
```
berendt, line 55: reference implementation? If Reference Implemention is correct it has to be changed in line 21.

Kubernetes-as-a-Services -> Kubernetes as a Service

docs/index.mdx
```diff
@@ -16,27 +16,46 @@ SCS is built, backed, and operated by an active open-source community worldwide.
...
+is provided by [Cluster-Stacks](https://docs-staging.scs.community/docs/category/cluster-stacks)
```
berendt, line 56: Cluster Stacks

berendt: I think this should be `/docs/category/cluster-stacks/`.

berendt (changes requested):

docs/index.mdx
garloff: Hmm, while I agree `in a` being lowercase, I consider the term Cloud-in-a-Box a single term rather than four words ... thus my preference to leave the hyphens in it. I believe we have used the form with hyphens at most places in the past. If we want to change that, we should probably grep through our repos and address it everywhere ...

garloff: Partially addressed by 9fc2e2fe.

docs/index.mdx
garloff: Done, see 9fc2e2fe.

docs/index.mdx
garloff: Agree, thanks!
Addressed in 9fc2e2fe.

docs/index.mdx
garloff: I capitalized Reference Implementation at both places now, giving it the property of a name. See 9fc2e2fe.

garloff: I have not removed the hyphens from Kubernetes-as-a-Service ... I consider this a single word.

docs/index.mdx
garloff: Done, see 9fc2e2fe.

garloff:
Thanks for your diligent review, @berendt !
I have addressed most of your suggestions, though not all.
@dirklossack always wanted me to use more hyphens than I did, you seem to be arguing for fewer dashes ...
My personal taste is that when using well-defined terms such as CiaB, KaaS, ... I tend to consider them one word, which is why I run them together into one by using hyphens. I'm certainly not the only one doing this, but I'm happy to learn from folks with better English skills whether or not this is common or a no-go (or both ;-))

PR #294, opened 2025-01-24
KaaS v2 (Cluster Stacks) has fully replaced v1.

garloff:
The wording was probably a year old, still talking about previews ...


matofeder (approved):
LGTM

PR #299, opened 2025-02-17
improve the description of the mailinglists

fkr:


garloff (commented):
I think we should mention that there are two ways to get on to the open lists.
(a) You are just curious on the communication here. Read the list archives, subscribe yourself to the mailing list at [mailman3 frontend](https://scs.sovereignit.de/mailman3/postorius/lists/).
(b) You like what the SCS community does and intend to interact with the community repeatedly. => Contact the SCS Project Board <project-board@lists.scs.community> to get onboarded onto the SCS nextcloud community instance. This will give you access to file sharing, polls, registry, ... and also automatically subscribe you to the mailing lists scs-tech, scs-member and the announce list.

garloff:
PS: If you agree we should document this, I can push a text proposal to the branch -- did not want to do this without discussion.

garloff (approved):
OK, let's keep it simple.

PR #305, opened 2025-03-18
Remove Cluster Stack Provider OpenStack

jschoone:
We plan to deprecate the CSPO from R8 in favor of the image kind of the OpenStack Resource Controller (ORC).
For archived docs, the CSPO repo can be mentioned.
Should me merge not before `images.openstack.k-orc.cloud` has been added to the Cluster Stacks.


garloff:
Jan, are the prerequisites fulfilled?
It would see so to me, but better double-check ...

garloff (approved):
We can merge this, no?

PR #312, opened 2025-04-30
feat: Introduce zuul job for building docs

gtema:


gtema:
recheck

gtema:
recheck

gtema:
@garloff Now the build looks okay with a workaround proposed in https://github.com/facebook/docusaurus/issues/448#issuecomment-908777029 Replacing the baseURL with the Zuul url is not trivial, since the URL at which document will be published may not be known until the moment of upload.
Other problem I have noticed during this work is https://github.com/SovereignCloudStack/docs/issues/314 where all documents pulled from somewhere else introduce broken links which cannot be caught. The "edit this page" functionality is best to be disabled as such due to that, but there might be another broken links.

PR #316, opened 2025-05-26
Move docs build forward 5 hrs

mbuechse:
From 0 am UTC to 5 am UTC, which is 7 am CEST.

Vastly increases the chance that we catch the latest nightly results from the compliance checks, in particular when they are moved back from 3 am UTC to 0 am UTC, as proposed in a concurrent PR: https://github.com/SovereignCloudStack/zuul-scs-jobs/pull/49

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -3,7 +3,7 @@ name: Build and Deploy to Github Pages
 on:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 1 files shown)


mbuechse:
@berendt We should consider this, but this opens quite the can of worms (because what we want to build on demand is only/mainly the results page, and other pages may have other requirements). In the meantime, is there any reason NOT to move the build forward?

fkr (approved):

PR #317, opened 2025-06-18
Feat/add jitsi help

garloff:
More information on our VC tool

### DIFF
--- a/community/cloud-resources/cloud-resources.md
+++ b/community/cloud-resources/cloud-resources.md
@@ -24,11 +24,11 @@ As suggested in [#155](https://github.com/SovereignCloudStack/standards/issues/1
 
 Example:
 
-| github handle | plusserver login    |
-| :-----------: | :-----------------: |
-|  frosty-geek  | u500924-frosty-geek |
-|      fkr      |     u500924-fkr     |
-|               |                     |
+|github handle | plusserver login    |
+|:-----------: | :-----------------: |
+| frosty-geek  | u500924-frosty-geek |
+|     fkr      |     u500924-fkr     |
+|              |                     |
 
 ### SCS2 Service Users
 


community/tools/jitsi.md
```diff
@@ -1,28 +1,138 @@
...
+| SCS-OSISM       | 614            |
+| SCS-Project     | 615            |
+| SCS-Forum       | 616            |
+| SCS-Kurt        | 617            |
+| SCS-Taskforce   | 618            |
+| SCS-ProjectTeam | 619            |
+
+Dial +49-221-292772-Suffix to connect.
+
+Rooms protected with a PIN would use 60x instead of 61x as suffix.
```
depressiveRobot, line 72: This could be explained in more detail. What does the dail-in with and without a PIN look like?

community/tools/jitsi.md
```diff
@@ -1,28 +1,138 @@
...
 Links to the meeting room (as well as dial-in information) are in the appointments
 in the public calendar.
 
 ## Usage
 
-Connect with a desktop browser (Chrome/Chromium or other blink based browser
-recommended due to superior WebRTC implementation with SimulCast/SVC for VP8/VP9 --
-Safari & Firefox work, but cause higher data traffic). For mobile devices use
-the Jitsi Meet App.
+Connect with a desktop browser or (for mobile devices) the Jitsi Meet App.
```
depressiveRobot, line 21: There is even a desktop app: https://github.com/jitsi/jitsi-meet-electron

mbuechse: Interesting. I tend to avoid Electron apps, because they seem to be more taxing on my PC than the regular web apps...

community/tools/jitsi.md
garloff: We have never started to use PINs. So I would actually have to retest this before documenting. It's more  ...like a hint that we could use PINs and are prepared to still have dial-in ...
(The way it is supposed to work by the way I have set up `extensions.conf`: You dial +49-221-292772-605 to connect into a conference in the SCS-Project room where the moderator has enabled PIN protection. You would then have a voice menu that asks you to enter the PIN which you do and then you are connected. I have tested this a lonnnnnng time ago.)

Feel free to test and add language.

Otherwise, my suggestion would be to leave as is and extend the documentation if there are calls where we actually want to use this.

community/tools/jitsi.md
garloff: OK, we can of course hint to it, although I am personally not a big fan of electron apps either.

depressiveRobot: I personally use it and others may be happy to get the hint, too.

PR #318, opened 2025-07-01
request access to SCS2 for depressiveRobot

depressiveRobot:
In order to develop on the SCS Compliance Check Pipeline and the SCS Compliance Monitor, I need credentials for SCS2 so that I can test against it.


fkr:
@costelter @neuroserve @marcvorwerk please assist :)

costelter (approved):
User created, added to project

PR #366, opened 2026-05-09
Feat/node 22

garloff:
Can we switch to node.js v22?


garloff:
This is not required for fixing the builds; PR #367 is enough for this.
Maybe still a good idea to go to node v22 as some of the used modules already complain.

community/tools/zuul.md
```diff
@@ -8,16 +8,18 @@ use Zuul as our main pipeline solution.
...
+  - Here is [an example](https://github.com/SovereignCloudStack/zuul-config/)
   - You can have a job section containing _self-defined_ jobs which you need to write on your own
   - You have to have a project section containing
     - the default-branch name
     - the merge-mode which should be used to auto-merge
     - the jobs to run in each pipeline (gh_check, gh_gate, gh_post, gh_tag)
-    - these pipelines are triggered by events which can be looked up [here](https://github.com/SovereignCloudStack/zuul_config/blob/main/zuul.d/gh_pipelines.yaml)
-    - some default jobs can be found [here](https://opendev.org/zuul/zuul-jobs/src/branch/master/playbooks)
+    - these [pipelines](https://github.com/SovereignCloudStack/zuul-config/blob/main/zuul.d/) are triggered by events
+    - ihere are [some default jobs](https://opendev.org/zuul/zuul-jobs/src/branch/master/playbooks)
```
depressiveRobot:
```suggestion
    - here are [some default jobs](https://opendev.org/zuul/zuul-jobs/src/branch/master/playbooks)
```

small typo

mbuechse:
> Actions will be forced to run with Node.js 24 by default starting June 16th, 2026.

@garloff You're right that we should do something, since Github is forcing us, and I'm sorry that I ignored this for so long out of complacency. However, it seems like they would prefer at least node 24. Do you think we could also leap to 24?

garloff:
If #379 works, we can drop this one, yes.

garloff:
Superceded by #379.

PR #367, opened 2026-05-09
Fix/link text descriptive

garloff:
We want to to to .nvmrc v20, but it complains about the here links, so fix them.


garloff:
We need this one to unbreak our CI.
.nvmrc v18 throws up on a RegExp.
With v20 that issue is gone, but now the link checker complains on links with here as text.
Fixing these turns up a few broken links.
Which I addressed as well.

So please approve.

fkr (approved):
thanks @garloff

PR #373, opened 2026-06-01
Update 2026-05-28-keystone.md

mrhopeman:
Update link for yaook security advisory

PR #374, opened 2026-06-01
Fix link to yaook security advisory.

garloff:
Thanks, @horazont!


garloff:
Already addressed in #373

PR #379, opened 2026-07-05
Upgrade node from v20 to v24.

garloff:
As mentioned in #366, github basically forces us to do so. The intermediate step to v22 may not be that insightful, so we may go to v24 directly.


depressiveRobot (approved):
Looks good. Tested it locally, no errors. I can start and browse it.

garloff:
Thanks, Marvin!

mbuechse:
Github still complains that the versions we use of certain actions are actually targeting node v20, so we should also bump those versions. *sigh*

PR #380, opened 2026-07-05
Feat/add osism concepts

garloff:
This was lacking from SCS docu, leaving out important pieces of the OSISM documentation.


garloff:
You can review the effects of this on https://docs-staging.scs.community/ ...

jklare:
Mh, I am a bit confused why half of the OSISM documentation should be (and already is) mirrored to the scs community docs page. Wouldn't it make more sense to write a short intro about what OSISM is and then just link to the OSISM docs page? This way, it would be a lot easier to ensure that the content is complete, up to date and in the right context.

garloff:
We do import most of the OSISM docs as the documentation for the SCS IaaS Reference implementation.
This is obviously a relic of the times when there was precisely one. So we have to evolve this conceptually:
- Allow for additional options and import the docs in here as well (if there is suitable material)
- Replace the whole thing by links

Until we have done restructuring, I would at least keep things up-to-date, which is what this patch was meant to do.

Apart from this, you are of course correct that we need to go through our docs and sort the production-ready things from the PoC things and the current docs from outdated docs (from unknown currency).

mbuechse:
I'm with Jan on this. Let's merge this by all means, but at the same time let's just link to the official docs (definitely don't wait for a grand overhaul of the whole thing).

jklare:
I opened a PR #386 that implements what was suggested and discussed here (as I understood it) for the OSISM part. Obviously, there are more things to do, and I would hope the community will pick those up eventually, but I would not want to wait for that before getting the restructuring started.

mbuechse:
I think this has been superseded by #386

PR #382, opened 2026-07-17
Support full clone URLs in getDocs (e.g. GitLab repos)

gerbsen:
getDocs.js hardcoded the GitHub host and prepended it to every `repo` entry in docs.package.json, so only GitHub repos could be imported. Resolve each entry to a clone URL instead: a full URL (https:// or git@) is used as-is, while a bare "owner/name" still falls back to the GitHub prefix for backwards compatibility.

This unblocks adding application examples that live on GitLab or other forges like codeberg.


fkr (approved):
LGTM!

fkr:
Indeed a very good addition. Thanks @gerbsen

PR #383, opened 2026-07-17
Speed up getDocs clones with shallow, single-branch fetch

gerbsen:
getDocs.js only needs a snapshot of each upstream repo's docs, but it did a full clone (entire history, all branches). Add `--depth 1` and `--single-branch` so each clone transfers only the latest commit of the default branch, cutting fetch time and bandwidth noticeably for the larger source repositories.

### DIFF
--- a/getDocs.js
+++ b/getDocs.js
@@ -14,8 +14,10 @@ const ghUrl = 'https://github.com/'
 repos.forEach((repo) => {
   const repoDir = `repo_to_be_edited/${repo.label}`
 
-  // Clone the repository
-  const cloneCommand = `git clone ${ghUrl + repo.repo} ${repoDir}`
+  // Clone the repository (shallow, single branch — we only need a snapshot of the docs)
+  const cloneCommand = `git clone --single-branch --depth 1 ${
+    ghUrl + repo.repo
+  } ${repoDir}`
   execSync(cloneCommand)
 
   // Remove git folders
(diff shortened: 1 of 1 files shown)


mbuechse:
This is a very welcome improvement!! Thanks! Can you update the branch so it can be merged?

getDocs.js
```diff
@@ -20,8 +20,8 @@ const cloneUrl = (repo) =>
 repos.forEach((repo) => {
   const repoDir = `repo_to_be_edited/${repo.label}`
 
-  // Clone the repository
-  const cloneCommand = `git clone ${cloneUrl(repo.repo)} ${repoDir}`
+  // Clone the repository (shallow, single branch — we only need a snapshot of the docs)
+  const cloneCommand = `git clone --single-branch --depth 1 ${cloneUrl(repo.repo)} ${repoDir}`
```
mbuechse:
```suggestion
  const cloneCommand = `git clone --depth 1 ${cloneUrl(repo.repo)} ${repoDir}`
```

quoting man page ("implies --single-branch")

```
       --depth=<depth>
           Create a shallow clone with a history truncated to the specified
           number of commits. Implies --single-branch unless
           --no-single-branch is given to fetch the histories near the tips of
           all branches. If you want to clone submodules shallowly, also pass
           --shallow-submodules.
```

mbuechse (approved):
Generally approve, with a suggestion for simplification. Out of curiosity: what role did Opus play in this very simple change?

getDocs.js
gerbsen: I have entered a CLI command and it generated:

- a new branch
- the code change with commit and description
- the pull request with description

while I already created the same thing for gitlab repo docs integration. Did not had to change any windows and the documentation is much better as If I would have done it myself.

PR #389, opened 2026-07-24
docs: remove orphaned placeholder pages

jklare:
## What

Removes 17 dead documentation pages that were carried over from the `docs-page` / `scs-docs` subtree merges but never wired into the consolidated sidebar. Each contained only a heading plus a `TODO` placeholder and was unreachable — no sidebar entry, no inbound links, and not covered by any autogenerated directory.

Deleted trees:
- `docs/02-iaas/overview/` — architecture, compute, knowledge, network, storage
- `docs/03-container/overview/` — architecture, knowledge
- `docs/03-container/deployment-examples/a/` — index, hardware, software
- `docs/03-container/guides/guide1.md`
- `docs/04-operating-scs/overview.md` and the `02-monitoring`, `03-incident-management`, `04-audits`, `05-lifecycle-management`, `06-logging` index pages

## Sidebar cleanup

`sidebarsDocs.js` held dangling references to the deleted files:
- The Container Layer **Overview**, **Deployment Examples** and **Guides** categories contained nothing but commented-out entries pointing at the deleted pages, so they would have rendered as empty `generated-index` pages — removed. Container Layer now resolves directly to **Components**, its only populated subsection.
- The commented-out `operating-scs/overview` entry was removed.

## Impact

No real content is removed and no live link or sidebar entry referenced any of these files, so the build is unaffected.

## Not included

Three orphaned pages that contain **real content** were intentionally left in place, as they likely want re-linking rather than deletion: `docs/01-getting-started/preinstall-checklist.md`, `docs/05-iam/iaas-roles.md`, and `docs/07-standards/index.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #390, opened 2026-07-24
docs: clean up remaining orphaned pages and re-link real ones

jklare:
Follow-up to #389, which removed the first batch of orphaned placeholder pages. This finishes the cleanup of the pages carried over from the `docs-page` / `scs-docs` subtree merges that were never wired into the consolidated sidebar.

## Remove remaining `TODO` stubs

Four more pages that were unreachable (no sidebar entry, no inbound links, not under any autogenerated directory) and contained only a heading plus a `TODO` marker:

- `docs/07-standards/index.md`
- `docs/01-getting-started/containerization.md`
- `docs/01-getting-started/overview.md`
- `docs/01-getting-started/virtualization.md`

## Re-link real content instead of deleting

Two orphaned pages held actual content, so they are wired into the sidebar rather than removed:

- `docs/01-getting-started/preinstall-checklist.md` — added as a top-level entry right after the landing page, as pre-installation planning material.
- `docs/05-iam/iaas-roles.md` — added under the Identity and Access Management category, alongside the federation setup page, as the reference for the IaaS-layer roles an SCS-compliant cloud offers.

## Impact

No live link or sidebar entry referenced the deleted stubs, and the two re-linked pages already existed on disk, so the build is unaffected. The sidebar module still parses cleanly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/01-getting-started/containerization.md
+++ b/docs/01-getting-started/containerization.md
@@ -1,7 +1,0 @@
----
-sidebar: 3
----
-
-# Containerization
-
-TODO
--- a/docs/01-getting-started/overview.md
+++ b/docs/01-getting-started/overview.md
@@ -1,7 +1,0 @@
----
-sidebar: 1
----
-
-# Overview
-
-TODO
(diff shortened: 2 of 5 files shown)


mbuechse (approved):
LGTM! Thanks ❤️

PR #392, opened 2026-08-05
Fix/docusaurus build warnings

gerbsen:
I was a little annoyed by all the warnings once you start the server, so I thought it might be a good idea to get rid of them. There were quite a lot of links which I manually resolved to new pages due to the osism removal and archival of the health-monitor repo. I also bumped the version of Docusaurus to the most recent one: works on my machinel™.

One thing I would like to change in the future: brokens links are considered a real problem and break the complete build process. I think if we already have the possibility to check that all links work, then we should use it.

### DIFF
--- a/docs.package.json
+++ b/docs.package.json
@@ -54,17 +54,6 @@
     "label": "openstack-health-monitor"
   },
   {
-    "repo": "SovereignCloudStack/scs-health-monitor",
-    "source": [
-      "docs/overview.md",
-      "docs/Workflow.md",
-      "docs/Testflow.md",
-      "docs/ObservabilityStack/SetupObservabilityStack.md"
-    ],
-    "target": "docs/04-operating-scs/components",
-    "label": "scs-health-monitor"
-  },
-  {
     "repo": "SovereignCloudStack/central-api",
     "source": "docs/*.md",
     "target": "docs/04-operating-scs/components",
--- a/docs.package.json
+++ b/docs.package.json
@@ -54,6 +54,17 @@
     "label": "openstack-health-monitor"
   },
   {
+    "repo": "SovereignCloudStack/scs-health-monitor",
+    "source": [
+      "docs/overview.md",
+      "docs/Workflow.md",
+      "docs/Testflow.md",
+      "docs/ObservabilityStack/SetupObservabilityStack.md"
+    ],
+    "target": "docs/04-operating-scs/components",
+    "label": "scs-health-monitor"
+  },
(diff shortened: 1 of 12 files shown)


docs.package.json
```diff
@@ -53,17 +53,6 @@
...
-    "repo": "SovereignCloudStack/scs-health-monitor",
-    "source": [
-      "docs/overview.md",
-      "docs/Workflow.md",
-      "docs/Testflow.md",
-      "docs/ObservabilityStack/SetupObservabilityStack.md"
-    ],
-    "target": "docs/04-operating-scs/components",
-    "label": "scs-health-monitor"
-  },
```
mbuechse, line 66: This is necessary.

gerbsen: Still don't understand why this does not happened locally after: 

```
npm run clear
npm run build
```

mbuechse: Well, apparently `npm run clear` (== `docusaurus clear`) won't delete the cloned repos from `getDocs.js`...

docusaurus.config.js
```diff
@@ -120,7 +122,18 @@ const config = {
...
         path: 'standards',
         routeBasePath: 'standards',
-        sidebarPath: require.resolve('./sidebarsStandards.js')
+        sidebarPath: require.resolve('./sidebarsStandards.js'),
+        // The vendored standards content (see getDocs.js) ships a README.md
+        // that Docusaurus would render at `/standards/` — the same route as our
+        // own standards/index.md. Skip it; index.md is the page we want there.
+        // Repeats Docusaurus' default excludes, which this option replaces.
+        exclude: [
+          'README.md',
```
mbuechse, line 131: Why was this not necessary before?

Or is this a preliminary step before we alter getDocs.js to not delete README.md?

mbuechse:
See https://docs-staging.scs.community/blog/

Blog posts are now truncated in the listing. I think this is an improvement. However, it's a change, so let's ask @garloff and @fkr whether it's okay.

gerbsen:
I'm also in the camp "improvement".

docusaurus.config.js
gerbsen: The `npm start` said very explicitly: 

```
[WARNING] Duplicate routes found!
- Attempting to create page at /standards/, but a page already exists at this route.
This could lead to non-deterministic routing behavior.
```

depressiveRobot: I don't get this error with "main".

PR #395, opened 2026-08-20
redirect old scope pages to new scope standards

depressiveRobot:
This will fix the broken links to the old scope pages (HTTP 404) by redirecting to the new scope standards. We should not have broken links, see ["Cool URIs don’t change"](https://www.w3.org/Provider/Style/URI.html) for a reasoning.

This is caused by SovereignCloudStack/standards#1199.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -72,12 +72,12 @@ const config = {
       {
         redirects: [
           {
-            to: '/standards/scopes/scs-0501',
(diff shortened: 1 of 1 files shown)


fkr (approved):
LGTM!

gerbsen:
High-level complaining but the order is a bit odd, I would switch from and to. :)

mbuechse:
Great work. Thanks!

